PR #939, opened 2015-03-23
add Travis CI

daxxog:
This change adds [Travis CI](https://travis-ci.org/) to enigma-dev. I have it set up to python install.py && make on both clang and gcc systems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: cpp
+sudo: false
+compiler:
+  - clang
+  - gcc
+install: python install.py
+script: make
--- a/CompilerSource/Makefile
+++ b/CompilerSource/Makefile
@@ -29,7 +29,7 @@ endif
 ###########
 
 CXX := g++
-CXXFLAGS += -std=c++11 -Wall -g -I./JDI/src
+CXXFLAGS += -std=c++0x -Wall -g -I./JDI/src
 LDFLAGS += -shared
 
 SOURCES := $(shell find . -name "*.cpp" -and ! -name "standalone_*")


TheExDeus:
Why did you change -std=c++11 into -std=c++0x? As far as I know they should be identical in most compilers, but as both GCC and Clang supports final c++11, then there is not reason to set draft version c++0x. Maybe I'm missing something.

Also, this adds Travis CI only to the ENIGMA compiler dll right? We should also have it for ENIGMA engine as well, but that would need a lot more complicated build-script (probably the cli would be used for this).

daxxog:
I changed -std=c++11 to -std=c++0x due to travis running older versions of gcc and clang.

I believe this is only testing building the dynamic library. All it does right now is run these commands:
```bash
python install.py
make
```

If we want to test other aspects of ENIGMA, more complicated build scripts will be needed. Maybe something like this:
```bash
make test
```

TheExDeus:
Here is some topics on how to get travis to use newer gcc.
http://gearchicken.com/wordpress/?p=105
http://stackoverflow.com/questions/22111549/travis-ci-with-clang-3-4-and-c11

Maybe we should test those. We also started to use C++11 quite extensively in the engine now and so it's possible that we can't even compile it with C++0x.

For more complicated tests we probably need to CLI (http://enigma-dev.org/forums/index.php?topic=2047.0). This is a separate executable inside the enigma-dev\CommandLine folder. This shows how to use it (http://enigma-dev.org/forums/index.php?topic=2047.msg20613#msg20613). It's not complete though, so many things we still cannot test. But the idea is that we could create some .egm projects that need to be compiled and executed to test ENIGMA engine. That way this would work as regression testing.

sorlok:
Note that c++0x and c++11 are exactly the same thing. The risk with c++0x is that (assuming it's gcc 4.6) it doesn't support some features of c++0x/11 that we might need. So I'm in favor of upgrading the compiler, I just want to point out that the two flags should have the same effect.

TheExDeus:
They should be the same, but I haven't explicitly seen a statement like that. As far as I know the 0x can refer to the draft version, while the 11 one refers to the final.

sorlok:
As far as I know, c++0x only refers to the drat version on very old compilers (where c++11 doesn't exist as a flag). 

The gcc docs do not say it explicitly, however (and I agree that c++11 is preferred).

daxxog:
@TheExDeus Personally I don't like the "hacky" way to get travis to use a newer gcc or clang because it requires using sudo. At the same time, using  c++0x is kinda hacky too. I will create a new branch with the ppa method of updating gcc/clang and see how it goes.

daxxog:
See #940 for a non c++0x method.

TheExDeus:
The C++11 solution was pulled, so this is no longer necessary.

PR #940, opened 2015-03-30
add Travis CI (travishelper.sh)

daxxog:
This uses a hacky sudo method with a bash script to set up travis with c++11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: cpp
+compiler:
+  - gcc
+  - clang
+before_install:
+  - sudo bash travishelper.sh $CXX
+install:
+  - python install.py
+script: make
--- a/travishelper.sh
+++ b/travishelper.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# - travishelper.sh
+# - helps set up travis with the proper gcc/clang
+
+echo CXX: $1
+
+#remove postgresql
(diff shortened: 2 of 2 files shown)


TheExDeus:
I don't get some things:
1) Why does sql needs to be removed? There is some conflict otherwise?
2) Why you get gcc from rayflood repo instead of just the regular one? Wouldn't apt-get update install the newer one?

Like this is a shorter version which apparently works (I haven't tested it):
```
   language: cpp
   compiler: g++

   before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
   - sudo apt-get update -qq

   install:
   - sudo apt-get install -qq g++-4.8
   - export CXX="g++-4.8"
```

daxxog:
@TheExDeus 
1) I removed SQL due to a versioning conflict causing apt to crash on "dist-upgrade"
2) I used rayflood repo because my travis script tests clang and gcc. That repo was the best one I could find for clang 3.5. Rayflood clang depends on rayflood gcc, thus I used both of those repos. If you can find a better one, let me know and I can change it.

I decided to go with a more heavy duty bash script because of testing both clang and gcc. My script takes the CXX env variable and conditionally installs the correct software and sets it up.

TheExDeus:
Alright then. All of this can also be changed at a later date if needed, so it's okay if we download the compilers from different branches. I'm sure there is a way to install newest clang and llvm from an official source though (this maybe: http://llvm.org/apt/ ?).

But if this works, then I'm okay to pull this.

daxxog:
Yes I tried to get clang from the official source (http://llvm.org/apt/), but I wasn't able to get it to work. If someone want to try to get the official sources to work with travis, go for it. Otherwise pull this and I know for sure it will work.

TheExDeus:
I will pull this for now.

PR #941, opened 2015-04-05
Fixes keyboard_lastkey and keyboard_key for OS-X

sorlok:
Due to an oversight, keyboard_key and keyboard_lastkey were never hooked up on OS-X. This fixes that.


TheExDeus:
Looks good.

PR #952, opened 2015-12-05
Added UK spellings of 'color' for GM:S compatability

faissaloo:
Since GameMaker:Studio allows for both spellings of 'color' to be used (for some strange reason) I implemented this in Enigma to make importing GameMaker:Studio projects easier.


TheExDeus:
I don't think we will merge this - See #950.

Issue #957, opened 2016-02-27
The function int64() doesn't exist in ENIGMA.


faissaloo:

faissaloo:
Fixed in #959

PR #959, opened 2016-02-28
Added int64(), fixing #957

faissaloo:


faissaloo:
That should be everything it needs, should sort out #957 completely.

TheExDeus:
We only support GCC as the compiler for ENIGMA, so there is no need to check for MSVC. We use C++11 and many very standard specific things that MSVC doesn't support anyway, so I doubt we can compile on it. Clang is an option though.
So I think you can remove that #ifdef. Also, does #ifdef WINDOWS really checks for MSVC? I can't seem to find a reference anywhere to it, but I guess it is safe for GCC as it requires _WIN32 or something similar to check for windows.
If anybody has more input I'll listen. If everything's okay I will merge.

faissaloo:
Done.  
Also, I did some more research regarding the ifdef, apparently #ifdef WINDOWS should have been #ifdef _WIN32. But since we don't need it anymore it doesn't really matter, I'll just keep that in mind in future I guess.

TheExDeus:
Yeah, but #ifdef _WIN32 checks for windows _platform_, not windows visual c++ _compiler_. So if you put #ifdef _WIN32 there then ENIGMA wouldn't compile on Windows as we use GCC on Windows as well.
Anyway, I'll let the commit stay here until someone else comments. If not then I will merge in a few days.

faissaloo:
I was expecting 'a few days' to me 2 or 3 days... Not 14 lol

TheExDeus:
Sorry, forgot about this. As nobody else said anything I will merge.

JoshDreamland:
Oh; sorry. I wasn't notified on this. I'm fine with this change; only thing I would have said is that since there is no overload accepting `const variant&`/`const var&`, calling this with a GM variable will fail.

faissaloo:
I can fix that later today

faissaloo:
Do I need an overload for both const variant& and const var& or just const var&?

PR #961, opened 2016-03-14
Added overload for const var& in int64()

faissaloo:
As mentioned by @JoshDreamland in #959

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/mathnc.h
+++ b/ENIGMAsystem/SHELL/Universal_System/mathnc.h
@@ -40,6 +40,7 @@
 
 #include <cstdlib> // random()
 #include "scalar.h"
+#include "libEGMstd.h"
 
 #define M_PI		3.14159265358979323846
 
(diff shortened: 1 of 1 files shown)


faissaloo:
Wait a second I just realised I did something really stupid here, gimmie a bit

faissaloo:
There we go, not sure how you guys feel about including libEGMstd in mathnc though.

TheExDeus:
I think leave to_string. We actually shouldn't use toString anyway. It was before we started using C++11 and in it we have a robust to_string function. A lot of small things like these should be removed, changed and updated. So in this case I think your original code was fine and we should in fact delete libEGMstd.h/.cpp if it doesn't provide new overloads (taking into account places it is used of course).

faissaloo:
I can't leave the to_string() because it doesn't compile, it gives me:  
```
‘to_string’ was not declared in this scope
```

TheExDeus:
It wouldn't work here anyway, as std::to_string() is doesn't have overloads for our custom types "var" and "variant". Maybe it could only work because "var" can implicitly cast to something that to_string can work with, but I doubt it.

Anyway, std::to_string() is C++11 feature (which your GCC supports) and to use it you must include <string> header and specify namespace (std) if no "using namespace std" has been used. See here: http://en.cppreference.com/w/cpp/string/basic_string/to_string

JoshDreamland:
For efficiency reasons, you probably want something like this:

```C++
inline int64_t int64(const var& x) { return int64(*x); }
inline int64_t int64(const variant& x) {
  return x.type == enigma::vt_tstr ? int64((string) x) : int64((double) x);
}
```

faissaloo:
How's that?

JoshDreamland:
LGTM

(I'm assuming it builds; I didn't plug the code in locally before making that suggestion.)

faissaloo:
Yeah it works, tried it with a variable that holds a string and a variable that holds a real and it worked both times.

PR #977, opened 2016-06-30
Fix instance_number to handle collection constants

### LINKED ISSUE #976: instance_number() and instance_find() don't handle the all constant correctly
I'm using this code in a draw event and laid out a few objects around the screen.
```
var i;
for (i=0;i<instance_number(all);i++)
{
    var inst;
    inst=instance_find(all,i)
    draw_text(inst.x,inst.y,string(i))
}
```  
On GM: 
GM  
On ENIGMA:  
ENIGMA  
ENIGMA gets data from garbage because all is a negative value and it results in instance_number() giving a massive number and instance_find() giving an invalid id which in the end results in inst.y and inst.x being extremely large. I have tested instance_find() by itself with a number given as the second argument and the issue with it is still there.

JoshDreamland:
Request created in GitHub. Someone please verify this builds and runs, since I still haven't set up a build bot.

This request should fix #976.

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/instance.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/instance.cpp
@@ -140,8 +140,8 @@ int instance_number(int obj)
       return (bool) enigma::fetch_inst_iter_by_int(obj);
     case all:
       return enigma::instance_list.size();
-    case noone: case global: case local:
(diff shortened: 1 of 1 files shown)


faissaloo:
Fixes the issue, thanks.
screencap

PR #984, opened 2016-08-01
Updated gcc.ey for Mac OS X

jspayd:


faissaloo:
@jspayd discovered that the gcc.ey for OSX was incorrect, causing builds to fail, this corrected it for him. Thread here: http://enigma-dev.org/forums/index.php?topic=2699.msg26099

TheExDeus:
As I don't have a Mac I won't be able to test it, but I can merge this if others think it's okay.

faissaloo:
I think I might be able to test it, gimmie a sec.

faissaloo:
Just tested it, it's a-ok for merge!

PR #992, opened 2016-08-07
Fix destructive force-push

rpjohnst:
We lost some history due to a force push from an old copy of the repository, this brings back the missing commits.


JoshDreamland:
For posterity: this branch contains all orphaned commits from the d9746dfc6d557c4c22c8bd8cf24e02e3272eade0 tree, which was obtained from the closed pull requests list. It's possible changes were merged in after that, but at very least, we have the 2015 commits.

PR #998, opened 2016-08-13
Fix some warnings

JoshDreamland:
These warnings turned out to be harmless, but they're bad smells, and a lot of them are due to formatting problems, particularly tabs vs spaces. I also moved the `M_PI` constant to a header; it was defined everywhere, and getting a bit out of hand.


JoshDreamland:
Based on the fact that this caused Robert no build errors when he inadvertently tested it in #999, I'm going to merge it tomorrow if no one merges it in the meantime. If anyone else would like to test this, first, please do so before then. (I will definitely merge this Friday evening, even if I forget tomorrow.)

JoshDreamland:
For visual inspection of diffs like this, I recommend adding `?w=1` to the diff view URL.

PR #1051, opened 2017-05-03
Fixed Box2D bodies not being destroyed

faissaloo:
Despite calling ```b2d_body_delete()``` Box2D objects wouldn't be destroyed, all that would happen would be that the memory for the struct that encapsulates Box2D objects would be freed, which would mean you would end up with ghost bodies in a Box2D world (Box2D bodies not attached to an ENIGMA object).


faissaloo:
Could someone take a look at this?
Thanks @JoshDreamland @RobertBColton

JoshDreamland:
I originally left this for Robert because he wrote that particular system. Your change looks reasonable to me, though, so I'll go  ahead and pull it.

faissaloo:
No problem!

PR #1072, opened 2017-09-20
Added a couple directory functions, fixes, tweaks.

time-killer-games:
- Replaced directory_delete() with directory_destroy() for GM compatibility
- In PFfilemanip.h I declared directory_delete() for those who still want that
- directory_destroy() / directory_delete() can now delete non-empty directories
- Replaced some, (not all), of the WinAPI functions with portable solutions
- Added functions directory_rename() and directory_copy()

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
@@ -144,6 +144,13 @@ int file_rename(std::string oldname,std::string newname)
     std::replace(str1.begin(),str1.end(),'/','\\');
     std::replace(str2.begin(),str2.end(),'/','\\');
 
+    char rpath1[MAX_PATH];
+    char rpath2[MAX_PATH];
+    GetFullPathName((char *)str1.c_str(),MAX_PATH,rpath1,NULL);
+    GetFullPathName((char *)str2.c_str(),MAX_PATH,rpath2,NULL);
+    str1 = string(rpath1);
+    str2 = string(rpath2);
+
     if (str1 != str2)
     {
         struct stat sb1;
@@ -165,6 +172,13 @@ int file_copy(std::string fname,std::string newname)
     std::string str2(newname);
     std::replace(str1.begin(),str1.end(),'/','\\');
     std::replace(str2.begin(),str2.end(),'/','\\');
+
+    char rpath1[MAX_PATH];
+    char rpath2[MAX_PATH];
+    GetFullPathName((char *)str1.c_str(),MAX_PATH,rpath1,NULL);
+    GetFullPathName((char *)str2.c_str(),MAX_PATH,rpath2,NULL);
+    str1 = string(rpath1);
+    str2 = string(rpath2);
 
     if (str1 != str2)
     {
@@ -336,6 +350,13 @@ int directory_rename(std::string oldname,std::string newname)
         }
     }
 
+    char rpath1[MAX_PATH];
+    char rpath2[MAX_PATH];
+    GetFullPathName((char *)str1.c_str(),MAX_PATH,rpath1,NULL);
+    GetFullPathName((char *)str2.c_str(),MAX_PATH,rpath2,NULL);
+    str1 = string(rpath1);
+    str2 = string(rpath2);
+
     if (str1 != str2)
     {
         struct stat sb1;
@@ -344,7 +365,10 @@ int directory_rename(std::string oldname,std::string newname)
             S_ISDIR(sb1.st_mode) &&
             stat((char *)str2.c_str(),&sb2) != 0)
         {
-            return (rename((char *)str1.c_str(),(char *)str2.c_str()) == 0);
+            if (str1 != str2.substr(0,str1.length()))
+            {
+                return (rename((char *)str1.c_str(),(char *)str2.c_str()) == 0);
+            }
         }
     }
 
@@ -373,6 +397,13 @@ int directory_copy(std::string dname,std::string newname)
         }
     }
 
+    char rpath1[MAX_PATH];
+    char rpath2[MAX_PATH];
+    GetFullPathName((char *)str1.c_str(),MAX_PATH,rpath1,NULL);
+    GetFullPathName((char *)str2.c_str(),MAX_PATH,rpath2,NULL);
+    str1 = string(rpath1);
+    str2 = string(rpath2);
+
     if (str1 != str2)
     {
         struct stat sb1;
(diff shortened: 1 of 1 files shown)


faissaloo:
Just a note: please keep your changes to one pull request instead of creating a new one for each commit, it makes things messy.

time-killer-games:
@faissaloo Sorry, I tried to do that initially but I couldn't figure it out. I'll make sure I know how to do it before I create my next pull request.

time-killer-games:
Before someone merges this pull request, I have one quick change to make because I'm OCD. The code should work the same, I just want to replace a few string comparisons to use GetFullPathName() to make them more reliable. i.e. two strings aren't equal if one of them has an extra slash somewhere in the path and the orther one doesn't. GetFullPathName() will remove excess slashes, eliminating this problem. I'll post a comment here when I have that done.

time-killer-games:
Alright, I'm done. It's ready to be merged unless anyone feels otherwise. :)

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
```diff
@@ -33,6 +33,13 @@
 #include <time.h>
 #include <sys/stat.h>
 
+#include <fstream>
+#include <algorithm>
+#include <direct.h>
+#include <dirent.h>
```
JoshDreamland: These are POSIX headers... they aren't guaranteed to exist on Windows.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
```diff
@@ -107,92 +114,454 @@ void ini_section_delete(std::string section)
...
-    } else {
-        return 1;
-    }
+int file_exists(std::string fname)
+{
+    std::string str(fname);
+    std::replace(str.begin(),str.end(),'/','\\');
+
+    struct stat sb;
+    return (stat((char *)str.c_str(),&sb) == 0 &&
```
JoshDreamland: `stat` is a POSIX function; this method should stay the way it was (or just be rewritten as `return GetFileAttributes(fname.c_str()) != INVALID_FILE_ATTRIBUTES;`). Alternatively, Win32 defines `PathFileExists(file)` which does exactly what we want (though as we've been discussing, it may be better to use the unicode equivalent of that).

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
```diff
@@ -107,92 +114,454 @@ void ini_section_delete(std::string section)
...
+int file_delete(std::string fname)
+{
+    std::string str(fname);
+    std::replace(str.begin(),str.end(),'/','\\');
+
+    struct stat sb;
+    if (stat((char *)str.c_str(),&sb) == 0 &&
+        S_ISREG(sb.st_mode))
+    {
+        return (remove((char *)str.c_str()) == 0);
```
JoshDreamland: The `remove` function is also POSIX... which Windows is not. The way the function was was fine, though, again, you may want to use `DeleteFileW` and `MultiByteToWideChar` to allow UTF-8.

I'd also remove the redundant `break;` commands in the old function....

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
```diff
@@ -107,92 +114,454 @@ void ini_section_delete(std::string section)
...
+
+    if (proceed || basename((char *)str1.c_str()) != basename((char *)str2.c_str()))
+    {
+        struct stat sb1;
+        struct stat sb2;
+        if (stat((char *)str1.c_str(),&sb1) == 0 &&
+            S_ISREG(sb1.st_mode) &&
+            stat((char *)str2.c_str(),&sb2) != 0)
+        {
+            return (rename((char *)str1.c_str(),(char *)str2.c_str()) == 0);
```
JoshDreamland: Same comments regarding `stat`/`rename`. They're just not Windows functions. Same comments regarding `MoveFileA` vs `MoveFileW`, if unicode support is what you're after.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
```diff
@@ -107,92 +114,454 @@ void ini_section_delete(std::string section)
...
+
+        return (RemoveDirectory((char *)str.c_str()) == true);
+    }
+
+    return 0;
+}
+
+int directory_rename(std::string oldname,std::string newname)
+{
+    std::string str1(oldname);
```
JoshDreamland: Like in your other pull request, you're handling a lot of things here that Windows knows how to handle. It's not going to explode and delete all your files if you ask to move `my/directory/path` to `my/directory/path`, or even `my/directory/path` to `my\directory\path`. It knows how to deal with that. And also, we can't use `rename`, here. Just make a single call to `MoveFileA`/`MoveFileW`.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
```diff
@@ -107,92 +114,454 @@ void ini_section_delete(std::string section)
...
+    }
+
+    return 0;
 }
 
-int directory_delete(std::string dname) {
-  BOOL result = RemoveDirectory(dname.c_str());
-  if (result) return 1;
-  else return 0;
+int directory_copy(std::string dname,std::string newname)
```
JoshDreamland: Windows spares us this complicated piece of logic (and other complicated pieces of logic in this file) via the [`IFileOperation`](https://msdn.microsoft.com/en-us/library/bb775771.aspx) interface. Anything in this file that you can't do simply using a single Win32 call, I recommend doing using that. If we want to support older versions of Windows, you may want to try using [`SHFileOperation`](https://msdn.microsoft.com/en-us/library/bb762164.aspx) instead. I could get behind using a macro test on the current header version to decide which one to use.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
time-killer-games: Well, they come with MinGW, and that's what ENIGMA uses to compile on Windows. Or does that not matter? i.e. the header may try to implement features that won't be recognized on some Windows systems?

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
time-killer-games: lol once upon a time my Windows code did use SHFileOperation. So yeah it won't be too hard to add that back in there fortunately.

JoshDreamland:
Can you merge this with #1073 so we can commit them as a transaction?

time-killer-games:
@JoshDreamland  It says "Only those with write access to this repository can merge pull requests."

JoshDreamland:
This has gone kind of stale, so I'm going to close it. I will leave #1073 open for now; please merge these changes into that branch (`patch-2`) if you still want them merged. The build bot should work properly on it, now.

PR #1075, opened 2017-09-27
The Unix equivalent to my Windows filemanip commit

time-killer-games:
The Unix equivalent to my Windows filemanip commit

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
@@ -17,9 +17,9 @@
 
 #include <cstdio>
 #include <cstdlib>
-#include <fstream>
 #include <unistd.h>
 #include <sys/stat.h>
+#include <sys/wait.h>
 
 #include <string>
 #include "PFfilemanip.h"
@@ -48,16 +48,16 @@ int file_rename(string oldname,string newname)
 
 int file_copy(string fname,string newname)
 {
-    ifstream srce(fname.c_str(),ios::binary);
-    if (srce.tellg() < 0) return 0;
-
-    ofstream dest(newname.c_str(),ios::binary);
-    if (dest.tellp() < 0) return 0;
-
-    dest << srce.rdbuf();
-    if (srce.tellg()-dest.tellp() != 0) return 0;
-
-    return 1;
+  pid_t c = fork();
+  if (c == -1) return false;
+  if (c) {
+    int status;
+    waitpid(c,&status,0);
+    return !status;
+  } else {
+    execlp("cp","cp",fname.c_str(),newname.c_str(),(char *)0);
+    exit(1);
+  }
 }
 
 int directory_exists(string dname)


time-killer-games:
Crap. I forgot to use git instead of the website. Sorry for making it a separate pull request!

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -20,48 +20,626 @@
...
+    return (stat((char *)str.c_str(),&sb) == 0 &&
+        S_ISREG(sb.st_mode));
+}
+
+int file_delete(std::string fname)
+{
+    std::string str(fname);
+
+    struct stat sb;
+    if (stat((char *)str.c_str(),&sb) == 0 &&
```
JoshDreamland: Why are you doing a stat on the file before deleting it? The `remove` method will set `errno` to tell you why the remove failed. But since you're returning false if the file doesn't exist, anyway, I'm not sure why you're bothering to check. This entire method is equivalent to `return remove(fname.c_str()) == 0;`.

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -20,48 +20,626 @@
...
+    {
+        return (remove((char *)str.c_str()) == 0);
+    }
+
+    return 0;
+}
+
+int file_rename(std::string oldname,std::string newname)
+{
+    std::string str1(oldname);
```
JoshDreamland: You're implementing a lot of logic here that the operating system handles for you. This is all well-defined behavior of `rename()`.

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -20,48 +20,626 @@
...
+    }
+
+    return 0;
 }
 
-int directory_create(string dname) {
-  return mkdir(dname.c_str(),S_IRUSR|S_IWUSR|S_IXUSR);
+int directory_rename(std::string oldname,std::string newname)
+{
+    std::string str1(oldname);
```
JoshDreamland: Once again, this is a ***LOT*** of code to do something as simple as rename a file. Most of the behavior you're defining here is already defined by `rename()`. If you want to have special handling for when the destination file already exists, that's fine, but that's as simple as calling `stat()`. Why are you trying to decompose these paths yourself? If you *absolutely need* that information, please use a call to `filename_name()` rather than re-implementing it yourself in every function. Regardless of what you do, please use more helpers instead of duplicating the code everywhere—it creates a maintenance problem, because someone will go in and make one fix to one function, but won't make the same change in the other dozen functions that are doing the same thing.

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -20,48 +20,626 @@
...
+    }
+
+    return 0;
 }
 
-int directory_delete(string dname) {
-  return rmdir(dname.c_str());
+int directory_copy(std::string dname,std::string newname)
+{
+    std::string str1(dname);
```
JoshDreamland: More helper functions, more FTW. This entire function should be a few lines in length. Something like this:

```C++
static std::string dst_root;
return ftw(src_root, [](const char* src_path, const struct stat* st, int typeflag) {
    std::string dst_path = dst_root + src_path;
    if (typeflag == FTW_D) return mkdir(dst_path.c_str(), st->st_mode);
    return !file_copy(src_path, dst_path);
}, 5);
```

If you need a special `file_copy` to handle hard links, that's fine, but I believe it would behoove our regular `file_copy` to handle links.

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -20,48 +20,626 @@
...
+    }
+
+    return 0;
 }
 
-int file_delete(string fname)
+int file_copy(std::string fname,std::string newname)
 {
-  return remove(fname.c_str());
+    std::string str1(fname);
```
JoshDreamland: Here and throughout: please extract path logic into helper functions, or use the existing ones (`filename_name`, `filename_path`, etc). This will greatly improve readability and maintainability. Copying a file is actually a hard problem if you want to avoid overwriting existing files and would like to handle links properly. I'm not sure why there are no POSIX shorthands for it.

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -20,48 +20,626 @@
...
+#include <stdlib.h>
+#include <limits.h>
+
 #include <string>
 #include "PFfilemanip.h"
 using namespace std;
 
 /* UNIX-ready port of file manipulation */
 
+struct dupslash
```
JoshDreamland: This is a nice gesture, but it's unnecessary. The filesystem doesn't care if there are duplicate slashes (and your code won't mind, either).

time-killer-games: I added that mainly to prevent directory_copy from infinitely copying inside itself.

Later on in the code you will see I do a substring comparison and if argument0 of directory_copy is a substring of argument1, then prevent the copy, otherwise it will infinitely copy itself inside itself with infinite subdirectories until you machince runs out of space. The two strings need to match, that's why I removed the duplicate slashes and used realpath(). I still have dupslash in there in case if realpath() fails, because yes I know realpath also removes extra slashes.

...or is that not an issue I should worry about catering to?

time-killer-games: Anyways let me know and I'll change it.

time-killer-games: Oh, I also did it for file_copy to make sure it returns 0 if argument0 = argument1, in which case will also need either dupslash, realpath, or both, (afaik, at least)

ENIGMAsystem/SHELL/Platforms/General/UNIXfilemanip.cpp
```diff
@@ -47,7 +48,16 @@ int file_rename(string oldname,string newname)
 
 int file_copy(string fname,string newname)
 {
-  return system(("cp "+fname+" "+newname).c_str()); // Hackish, but there's no good implementation on Linux
+    ifstream srce(fname.c_str(),ios::binary);
```
JoshDreamland: I hate to heckle you what you decided to keep, but this breaks link handling—`copy.c` is actually an incredibly complicated source file. You can take a look at it [in the coreutils source repository](http://git.savannah.gnu.org/cgit/coreutils.git/tree/src/copy.c#n1844). Look at that monster. Some thousand lines go into determining how to perform the copy operation we're requesting.

That said, the old function is a big security hole; I'm not sure if I wrote it or just approved it, but there are better ways to invoke `cp` than using the `system()` routine. You could do a fork-and-exec pattern to invoke it, like this:

```C++
pid_t c = fork();
if (c == -1) return false;
if (c) {
  int status;
  waitpid(c, &status, 0);
  return !status;
} else {
  execlp("cp", fname.c_str(), newname.c_str());
  exit(1);
}
```

Care to try that out?

PR #1127, opened 2018-01-02
clang format on engine

fundies:
blind format on all files in engine. gimme list of which to completly exclude and which need minor fixes.


JoshDreamland:
Try just running this on the files from your other PR.

JoshDreamland:
Let's not do this.

PR #1131, opened 2018-01-07
Set _NET_WM_PID on X11 windows.

JoshDreamland:
This will enable a test harness to properly identify ENIGMA windows to manipulate them and read their attributes.


JoshDreamland:
As I said, I think formatting is a marvelous idea if you're editing the code already and are prepared to explain how the code you've formatted works. As one of the original authors of the code, I'm comfortable formatting it when I change it.

PR #1135, opened 2018-01-08
Command line Single-Object Game (.sog) support

JoshDreamland:
Adds simple game file support to the command line interface. This should enable a test harness to quickly launch and verify simple test cases.

Also fixes a resource leak I noticed in  X11 while writing tests.


CommandLine/emake/Main.cpp
```diff
@@ -2,6 +2,45 @@
 #include "EnigmaPlugin.hpp"
 #include "Game.hpp"
 
+#include <fstream>
```
fundies: Move this to it's own file 3FG.cpp

CommandLine/emake/Main.cpp
```diff
@@ -34,7 +71,37 @@ int main(int argc, char* argv[])
     bool _run = options.GetOption("run").as<bool>();
     if (!_run) plugin.HandleGameLaunch();
 
-    return plugin.BuildGame(game.ConstructGame(), mode, options.GetOption("output").as<std::string>().c_str());
+    Game game;
+    std::string input_file = options.GetOption("input").as<std::string>();
+    if (input_file.size()) {
+      if (input_file.size() < 4 || input_file[input_file.length() - 4] != '.') {
```
fundies: Just store extension in string and check if empty.

CommandLine/emake/Main.cpp
```diff
@@ -34,7 +71,37 @@ int main(int argc, char* argv[])
...
+        std::cerr << "Error: Unknown filetype: can't read input file \""
+                  << input_file << "\"." << std::endl;
+        return 1;
+      }
+      std::string ext = tolower(input_file.substr(input_file.length() - 3, 3));
```
fundies: not all files have 3 letters as extensions. use something like find_last_of(".") instead

CommandLine/emake/OptionsParser.cpp
```diff
@@ -90,7 +90,8 @@ OptionsParser::OptionsParser() : _desc("Options")
   _desc.add_options()
     ("help,h", "Print help messages")
     ("info,i", opt::value<std::string>(), "Provides a listing of Platforms, APIs and Extensions")
-    ("output,o", opt::value<std::string>()->required(), "Output file")
+    ("input",   opt::value<std::string>(), "Input game file")
```
fundies: hint at supported file types here

CommandLine/emake/Main.cpp
```diff
@@ -34,7 +71,37 @@ int main(int argc, char* argv[])
...
+      if (ext != "3fg") {
+        if (ext == "egm") {
+          std::cerr << "EGM format not yet supported. "
+                       "Please use LateralGM for the time being." << std::endl;
+        } else if (ext == "gmx") {
+          std::cerr << "GMX format not supported. "
+                       "Please use LateralGM or find a converter." << std::endl;
+          return 1;
+        } else {
+          std::cerr << "Error: Unknown filetype: can't read input file \""
```
fundies: unkown file type: %extension% ...

fundies:
This is also broken and throwing exceptions upon launch

CommandLine/emake/Game.hpp
```diff
@@ -9,22 +9,33 @@
...
+  std::vector<GmObject> _objects;
+  std::vector<Room> _rooms;
+
+  std::vector<std::vector<MainEvent>> _main_events;
+  std::vector<std::vector<Event>> _events;
+  std::vector<std::vector<Instance>> _instances;
+  std::vector<std::string> _strings;
+
+  const char* SS(std::string str);
+  void AddSimpleEvent(std::vector<MainEvent>* main_events,
```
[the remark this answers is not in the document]
fundies: I believe supporting SOG would be better for testing overall so we can ensure events like collisions, animation end and game end are actually being triggered properly in our test cases. That said, I think the number of events in SOG should be optional as to not make it a pain to create one because even  I have no idea what half the events are named in GM. I'm willing to accept 3fg but I believe we would be wanting sog instead soon...

.gitignore
```diff
@@ -46,5 +46,5 @@ ENIGMAsystem/SHELL/.objs
 MacOS/build/Release/EnigmaXcode.app/Contents/Info.plist
 MacOS/build/Release/EnigmaXcode.app/Contents/Resources/*
 *.stackdump
-emake
+/emake
```
fundies: why?

CommandLine/emake/3FG.hpp
```diff
@@ -0,0 +1,8 @@
+#ifndef EMAKE_3FG_HPP
+#define EMAKE_3FG_HPP
+#include <string>
```
fundies: our includes before system ones please

.gitignore
JoshDreamland: You were ignoring everything called "emake" (including the emake source folder). Now it just ignores the binary at the repo's root.

CommandLine/emake/Game.hpp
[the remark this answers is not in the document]
JoshDreamland: The idea is that we will have EGM before that. However, even in the presence of EGM, perhaps SOG will be useful. I've implemented it.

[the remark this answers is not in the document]
RobertBColton: Agree here, lots of quick indie competitions also end up as SOG. So it's a bonus that it might also be able to run semi-real games! Theoretically speaking, a single object is all you actually ever need in order to create any game in GM. While it may be really difficult and messy, you could, hypothetically, make something like Iji with a single object.

CompilerSource/compiler/event_reader/event_parser.h
```diff
@@ -53,6 +53,11 @@ enum p_type {
 #include <vector>
 #include <map>
 
+using std::map;
```
fundies: "using" in headers will leak into header including this header. I know this isn't as a big issue in the compiler but it is still a bad habbit.

CompilerSource/compiler/event_reader/event_parser.cpp
```diff
@@ -22,6 +22,7 @@
 #include <vector>
 #include <map>
 using namespace std;
+#define tostring to_string
```
fundies: is there really a need for this? why not just atually replace it...

CommandLine/emake/SOG.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <fstream>
+#include <iostream>
+
+bool ReadSOG(const std::string &input_file, Game *game) {
+  event_parse_resourcefile();
+  map<string, evpair> event_filenames;
+  for (evpair e : event_sequence) {
+    event_filenames[event_get_function_name(e.first, e.second) + ".edl"] = e;
+    if (!e.second && event_is_instance(e.first, e.second)) {
+      for (int i = 1; i < 127; ++i) {
```
fundies: What is 127 here? Please use constants when possible.

CommandLine/emake/SOG.hpp
```diff
@@ -1,8 +1,8 @@
 #ifndef EMAKE_3FG_HPP
 #define EMAKE_3FG_HPP
-#include <string>
 #include "Game.hpp"
+#include <string>
 
-bool read_3fg(const std::string &input_file, Game *game);
+bool ReadSOG(const std::string &input_file, Game *game);
```
fundies: nice catch on the snakecase. I missed that in orginal review.

CommandLine/emake/Game.hpp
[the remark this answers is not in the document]
fundies: I think SOG would be cleaner than egm for adding tests to our source tree.

CommandLine/emake/Main.cpp
fundies: I was more concerned with someone adding something like 3fgz support later.

CommandLine/emake/Main.cpp
fundies: How would you compute a mime on a folder?

CompilerSource/compiler/event_reader/event_parser.h
JoshDreamland: Cleaning up this header is not something I intend to do in this pull request. Attempting to remove this would create dozens of useless line changes.

CompilerSource/compiler/event_reader/event_parser.cpp
JoshDreamland: More useless line changes.

CommandLine/emake/SOG.cpp
JoshDreamland: The 127 is as arbitrary as it sounds. It needs to be bigger than the biggest vk_ constant, and big enough for alarms and user events, etc (which are eclipsed by the biggest vk_ constants). This is mostly a hack to work around the fact that the engine uses numbers instead of names for keys in keyboard events.

fundies: if you don't have a name for it atleast leave a comment as to why you put it there so someone else reading this can follow your logic.

CompilerSource/compiler/event_reader/event_parser.cpp
fundies: hardly, we have toString, to_string, tostring, and std::to_string in  our sources now some are the same func implemented multiple times, some are crappy macros like this and some are usally useful functions. Adding another macro only piles onto that mess and makes it harder to find the actual function being called.

CompilerSource/compiler/event_reader/event_parser.h
fundies: wasnt't asking that you clean it just that you don't dirty it further :)

JoshDreamland: As it stands (at HEAD), the file depends on its caller putting this data in the global namespace for it. As you have pointed out, this is a code smell. I have modified the file so that now it ensures its global namespace pollution itself. This improves the header by making it possible to include it without being charged with the task of polluting the global namespace on your own. 

Regardless, this is not going to be a bureaucratic process in which everyone modifying a file must feel obligated to "fix" the file to fit a more hip and modern understanding of how it should look. Or, as you'd argue, to "not dirty it further." I always point out to contributors when they are rubbing against something gross about the project, but I specifically tell them not to worry about fixing it, because that's how you discourage people from contributing to your codebase.

CommandLine/emake/SOG.cpp
JoshDreamland: That's fair; I will do so before I submit this, tonight.

CompilerSource/compiler/event_reader/event_parser.h
fundies: If it were anyone else I might be more leinent and just post a bug along with this commit until I or someone else cleaned up your mess. But, any addition to file is technically a modification and this is a file you created. There's really no need to add yet another macro for toString because you won't do a simple find and replace in fear of having to change 12 lines in a file that YOU created, lol. I know you'll commit this now, forget about it and we'll run into issue later. There's reasons we started reviewing code and not just blindy taking changes from others like Robert and myself, and this is one. Honestly, If it were a new contributer you would be the one helping mitigate the issue, thus you should be the one to fix it here or as I said earlier atleast not further it.

RobertBColton: I think you've both made excellent points here, you're just coming from different perspectives. @fundies is just focused on the here and the now.

PR #1179, opened 2018-02-05
Protocol Buffers for Compiler Communication

fundies:
makefiles are shit but should be good enough to get you started @RobertBColton


RobertBColton:
Don't start labeling things "Don't Merge" again like I used to do. There is no rush here, merging is not even possible since Travis is failing. Anyway, I'll take a look.

fundies:
Marked as that as its not even ready for review. this is fake pr to get you off your arse

RobertBColton:
Sure, just use the "Feedback needed" label for that or create a new one. I think GitHub actually has a setting somewhere to prevent merging a PR that has a specific label, would need to ask Josh about that though.

fundies:
took 17 tries but it builds on appveyor and travis now

RobertBColton:
I have a suggestion so far, you should change this:
```protobuf
optional int32 id = 2 [(gmx) = "GMX_DEPRECATED"]; // old
optional int32 id = 2 [(GMX_DEPRECATED)]; // new
```
Attributes work a little bit like command line flags, you can have one that's just a simple switch, either present or not present. That is how the built-in optional and repeated attributes work.

fundies:
you write 1 line yet?

RobertBColton:
@fundies smooth for paths is not deprecated, it's just called kind now

CommandLine/libGMX/gmx.cpp
```diff
@@ -0,0 +1,315 @@
...
+              break;
+            }
+            case google::protobuf::FieldDescriptor::CppType::CPPTYPE_FLOAT: {
+              refl->SetFloat(m, field,
+                             (isAttribute) ? attr.as_float() : (isSplit) ? std::stof(splitValue) : xmlValue.as_float());
+              break;
+            }
+            case google::protobuf::FieldDescriptor::CppType::CPPTYPE_BOOL: {
+              refl->SetBool(m, field,
+                            (isAttribute) ? attr.as_bool() : (isSplit) ? (std::stof(splitValue) != 0) : (xmlValue.as_int() != 0));
```
fundies: shouldn't be stof here

fundies:
mkay. I have no idea what 80% of crap in gmx is. So if I got anything else wrong its no suprise

CommandLine/libGMX/gmx.cpp
RobertBColton: why?

RobertBColton:
Ok, after a bit more cleanup I can say that actions to code is working almost completely now. Not only did I test all the permutations of actions using GMS, I also tested some games that had drag and drop problems in LGM.

One of these was the `JumpDemonstration1.gmk` from our forums, where users were having a problem with child actions getting the with scope of their parent question actions:
https://enigma-dev.org/forums/index.php?topic=2643.msg26695#msg26695

I had to modify the example slightly to draw an ellipse because I don't have sprites loading in the CLI yet. This edge case works correctly now in the CLI using my reserved `__if__` var inside a with scope.
Pre-JDI:
```gml
var __if__ = false;
with (object_Ball)
__if__ = action_if_variable(Jump,1,0)
if (__if__)
{
draw_set_color(mycolor);

/**/
}
with (object_Ball)
__if__ = action_if_variable(Jump,0,0)
if (__if__)
{
draw_set_color(c_yellow);

/**/
}
```
Post-JDI:
```cpp
variant enigma::OBJ_object_Arrow::myevent_draw()
{
  var __if__ = 0;
  with((object_Ball))
    __if__ = action_if_variable(enigma::varaccess_Jump(int(self)), 1, 0);
  if((__if__))
  {
    draw_set_color(mycolor);
    
  }
  with((object_Ball))
    __if__ = action_if_variable(enigma::varaccess_Jump(int(self)), 0, 0);
  if((__if__))
  {
    draw_set_color(c_yellow);
    
  }
  action_draw_rectangle(0, 0, 50, 50, 0);
  
  return 0;
}
```
So, yeah this works better than both LGM and GMS now! Only thing left is to add oneof for the action arguments so actions that refer to other resources work. I'll do that next.

RemoveRusky:
You still need to add the names to scripts and shaders in the gmxreader.

RobertBColton:
Lmao, no lol, I already had them in there and was wondering why it wasn't working. Check my last commit, the scripts and shaders do work now, I just built a GMX with them.

RemoveRusky:
The work but if you were to load them in the ide there is currently no names set for them

RobertBColton:
I literally just fixed the names to also remove the `.gml` and `.shader` extensions! ENIGMA can't write them if they either don't have a name, or name contains an extension, and will error. Can you not see de4789474f6bed8f18fff60d77e1338c22db336b?

RemoveRusky:
Sorry that wasn't showing up in my log. Also, have tou tried using setenum instead of setenumvalue?

RobertBColton:
Yes, originally you had it commented out, it takes an EnumDescriptor but I can't find any example online of how you are supposed to call it to set the value? If you can find that it would be helpful. I think you had it commented out though because you didn't know either.

RemoveRusky:
Nah I just didnt bother trying. Looks like its something like this

const EnumDescriptor * enumValue = desc->enum_type(<int here>); 
setenum(enumvalue....

RobertBColton:
Right, that's not helpful, how do you put the integer in there? I'm not writing a giant ass switch statement...

RemoveRusky:
it literally says "int here" u twat

RobertBColton:
There, it now looks like Travis is passing green. If the change was more complicated I would not be interested in supporting older users, but this is fine, we can do this.

RemoveRusky:
gmx dun yet?

RemoveRusky:
gmx dun yet?

RemoveRusky:
y u work on ide when this not dun

RemoveRusky:
anyways can we finish this now?

RemoveRusky:
dun yet?

RemoveRusky:
why u rename

RobertBColton:
Alright, I've added the oneofs to finish the action argument conversion, and the FPS example is now loading. However, inheritance is causing a segmentation fault when I fire the gun, just after I hear the start of the gunshot sound. Collision detection is otherwise working fine.

FPS example loaded in protobuf

RobertBColton:
Ok, apparently this always had a segfault which was hidden by older gcc, the fix is #1191.

RobertBColton:
Alright with object_is_ancestor fixed, and my room color fix, the FPS example works %100 now:
FPS Example Completely Working

RemoveRusky:
who uses bgr. rgb is mater race

RemoveRusky:
pretty sure i saw bgra being used in our engine not abgr

RobertBColton:
It's definitely ABGR:
https://github.com/enigma-dev/enigma-dev/blob/24ed2d765cad4316f1fc91c95f876cab32b81bcd/ENIGMAsystem/SHELL/Graphics_Systems/General/GScolors.cpp#L91

Again, red is always 255 as an int32.

RemoveRusky:
eitherway using bgra is dumb. everyone uses rgb

RobertBColton:
RIght, it's also weird because in your code, it looks like its RGB but the shifting makes it actually BGR. When we say RGB, we are talking about how it looks as a hex value:

0xRRGGBB

Just to clarify.

RemoveRusky:
so are we good to merge this? or did you want to add the other loading features we discussed?

RemoveRusky:
@RobertBColton I deleted duplicate lodepng for you but this isn't working on my end it fails to load the images (on your previous commit).  I think it's cause gmx uses \ and you never replaced with / but I'm too lazy to check

RobertBColton:
Well I don't have a Linux instance configured right now, so why don't you just investigate.

RemoveRusky:
[fps.zip](https://github.com/enigma-dev/enigma-dev/files/1763100/fps.zip)

RemoveRusky:
Audio is still failing but travis is listing it as green. It actually segfaults https://travis-ci.org/enigma-dev/enigma-dev/jobs/347286870#L2413 . Locally, theres no segfault but it does fail the sound_exists test

RemoveRusky:
So I fixed the openal errors in previous build and now travis displays same behavior I have locally https://travis-ci.org/enigma-dev/enigma-dev/jobs/347320482#L2436

RobertBColton:
Then why is it saying the test overall failed?

RemoveRusky:
it's saying it failed cause the test failed :/

RobertBColton:
I got bad news, looks like the latest version of GMS 1.4.1788 is totally fubar with sounds.

1) I create a new project.
2) I create a sound.
3) I import a random mp3 from on my file system using the load button the sound editor:
Sound Imported
4) I save the game, and this is what the sound's `<data>` and `<origname>` elements contain:
```xml
<origname>sound\audio\sound0.mp3</origname>
<data>sound0.mp3</data>
```
5) I rename the sound from `sound_0` to `sound_test`.
6) I hit the ok button and resave the game.
7) The directory now contains a copy of the old sound and the new sound:
Copy of both sound GMXs
8) The audio folder contains a copy of the mp3 with the old name but not the newer one:
MP3 has old name
9) Now here is the problem, the new `sound_test.sound.gmx` contains the new name of the file, which doesn't exist:
```xml
<origname>sound\audio\sound_test.mp3</origname>
<data>sound_test.mp3</data>
```

CommandLine/protos/resources/Event.proto
```diff
@@ -0,0 +1,83 @@
...
+    string path = 6;
+    string script = 7;
+    string object = 8;
+    string room = 9;
+    string font = 10;
+    string timeline = 11;
+  }
+}
+
+enum ActionKind {
```
JoshDreamland: Where did these come from? Are these  identical to the constants used in LGM?

CommandLine/protos/resources/Object.proto
```diff
@@ -0,0 +1,41 @@
...
+  repeated Event events = 10 [(gmx) = "events/event"];
+
+  enum PhysicsShape { CIRCLE = 0; BOX = 1; SHAPE = 2; }
+
+  message PhysicsShapePoint {
+    optional int32 x = 1 [(gmx) = "GMX_SPLIT/0"];
+    optional int32 y = 2 [(gmx) = "GMX_SPLIT/1"];
+  }
+
+  optional bool use_physics = 11 [(gmx) = "PhysicsObject"];
```
JoshDreamland: Do we have the ability to extract these into a nested message? I'd really like to.

CommandLine/protos/resources/GameInformation.proto
```diff
@@ -0,0 +1,22 @@
...
+package buffers.resources;
+
+import "options.proto";
+
+message GameInformation {
+  optional int32 background_color = 1;
+  optional bool embed_game_window = 2;
+  optional string form_caption = 3;
+
+  optional int32 left = 4;
```
JoshDreamland: Points for the most awkward ordering I can think of.

CommandLine/protos/resources/Path.proto
```diff
@@ -0,0 +1,25 @@
...
+  message Point {
+    optional int32 x = 1 [(gmx) = "GMX_SPLIT/0"];
+    optional int32 y = 2 [(gmx) = "GMX_SPLIT/1"];
+    optional int32 speed = 3 [(gmx) = "GMX_SPLIT/2"];
+  }
+
+  optional string name = 1;
+  optional int32 id = 2 [(gmx) = "GMX_DEPRECATED"];
+
+  optional string background_room_name = 3 [(gmx_id) = "backroom"];
```
JoshDreamland: Isn't this group just for the IDE? If we have the ability to extract these, I'd extract them.

CommandLine/protos/resources/Font.proto
```diff
@@ -0,0 +1,40 @@
...
+  optional int32 size = 4;
+  optional bool bold = 5;
+  optional bool italic = 6;
+
+  repeated Range ranges = 7 [(gmx) = "ranges/range0"];
+  repeated Glyph glyphs = 8 [(gmx) = "glyphs/glyph"];
+  
+  optional bool renderhq = 10;
+  optional uint32 charset = 11;
+  optional uint32 aa = 12;
```
JoshDreamland: I'd name this "antialias," I believe.

CommandLine/protos/resources/Font.proto
```diff
@@ -0,0 +1,40 @@
...
+  optional bool include_ttf = 13 [(gmx) = "includeTTF"];
+  optional string ttf_file = 14 [(gmx) = "TTFName"]; 
+  optional int32 texture_group = 15 [(gmx) = "texgroups/texgroup0"];
+  optional string image = 16;
```
JoshDreamland: Same as the other images.

CommandLine/protos/resources/Background.proto
```diff
@@ -0,0 +1,31 @@
...
+  optional int32 vertical_spacing   = 16 [(gmx) = "tilevsep"];
+
+  optional int32 h_tile = 17 [(gmx) = "HTile"];
+  optional int32 v_tile = 18 [(gmx) = "VTile"];
+  optional int32 texture_group = 19 [(gmx) = "TextureGroups/TextureGroup0"];
+  optional bool for3D = 20 [(gmx) = "For3D"];
+  optional uint32 width = 21;
+  optional uint32 height = 22;
+
+  optional string image = 3 [(gmx) = "data", (file_path) = true];
```
JoshDreamland: I believe the type you want is `bytes`. Also consider the Image submessage.

CommandLine/protos/resources/Sprite.proto
```diff
@@ -0,0 +1,41 @@
...
+
+  optional int32 type = 17;
+  optional int32 h_tile = 18 [(gmx) = "HTile"];
+  optional int32 v_tile = 19 [(gmx) = "VTile"];
+  optional int32 texture_group = 20 [(gmx) = "TextureGroups/TextureGroup0"];
+  optional bool for3D = 21 [(gmx) = "For3D"];
+  optional uint32 width = 22;
+  optional uint32 height = 23;
+
+  repeated string subimages = 3 [(gmx) = "frames/frame", (file_path) = true];
```
JoshDreamland: Same as the other images.

CommandLine/protos/resources/Sprite.proto
```diff
@@ -0,0 +1,41 @@
...
+  optional int32 bbox_left = 13;
+  optional int32 bbox_right = 14;
+  optional int32 bbox_top = 15;
+  optional int32 bbox_bottom = 16;
+
+  optional int32 type = 17;
+  optional int32 h_tile = 18 [(gmx) = "HTile"];
+  optional int32 v_tile = 19 [(gmx) = "VTile"];
+  optional int32 texture_group = 20 [(gmx) = "TextureGroups/TextureGroup0"];
+  optional bool for3D = 21 [(gmx) = "For3D"];
```
JoshDreamland: What on earth does this mean? Generate scales and filters? Could you comment this, and possibly use a more descriptive name?

CommandLine/protos/resources/Font.proto
```diff
@@ -0,0 +1,40 @@
...
+
+  repeated Range ranges = 7 [(gmx) = "ranges/range0"];
+  repeated Glyph glyphs = 8 [(gmx) = "glyphs/glyph"];
+  
+  optional bool renderhq = 10;
+  optional uint32 charset = 11;
+  optional uint32 aa = 12;
+  optional bool include_ttf = 13 [(gmx) = "includeTTF"];
+  optional string ttf_file = 14 [(gmx) = "TTFName"]; 
+  optional int32 texture_group = 15 [(gmx) = "texgroups/texgroup0"];
```
JoshDreamland: This is a texture atlas, I assume..? I didn't realize GM was quite this far along.

CommandLine/protos/game.proto
```diff
@@ -0,0 +1,27 @@
...
+import "resources/Background.proto";
+import "resources/Object.proto";
+import "resources/Script.proto";
+import "resources/Shader.proto";
+import "resources/Path.proto";
+import "resources/Font.proto";
+import "resources/Timeline.proto";
+import "resources/Room.proto";
+
+message Project {
```
JoshDreamland: This has to be the most amazing thing about this change.

CommandLine/libGMX/gmx.cpp
```diff
@@ -0,0 +1,382 @@
...
+    str.replace(pos, substr.length(), nstr);
+    pos += nstr.length();
+  }
+  return str;
+}
+
+inline std::string extract_file_path(std::string dir, std::string value) {
+  value = string_replace_all(value, "\\", "/");
+  const std::string end = dir.substr(dir.find_last_of("/") + 1);
+  if (value.find(end) == 0)
```
JoshDreamland: Why are you checking if `value` starts with the last path of `dir`? Are you assuming the next char will be `/`? Or do you expect the two to be equal? This function is doing more than just extracting a path.

CommandLine/libGMX/gmx.cpp
```diff
@@ -0,0 +1,382 @@
...
+#include <map>
+#include <sstream>
+#include <vector>
+#include <algorithm>
+
+#include "gmx.h"
+
+namespace gmx {
+
+inline std::string string_replace_all(std::string str, std::string substr, std::string nstr)
```
JoshDreamland: Put `namespace {}` around this static stuff.

CommandLine/libGMX/gmx.cpp
```diff
@@ -0,0 +1,382 @@
...
+        constants[node.parent().attribute("name").value()] = node.text().as_int();
+      } else {
+        resMap[std::string(node.parent().name()) + "s"].push_back(node.value());
+      }
+    }
+    return true;
+  }
+};
+
+std::string FileToString(const std::string &fName) {
```
JoshDreamland: I think we have like five of these elsewhere in the compiler... I guess another doesn't even hurt anything.

CommandLine/libGMX/gmx.cpp
```diff
@@ -0,0 +1,382 @@
...
+        if (child.empty()) attr = node.attribute(xmlElement.c_str());
+        isAttribute = !attr.empty();
+
+        if (child.empty() && !isAttribute && !field->is_repeated()) {
+          // ename only exists if etype = 4. Also, etype and enumb don't exist in timeline events
+          pugi::xml_attribute a = node.attribute("eventtype");
+          if (xmlElement != "ename" && a.as_int() != 4 && node.path() != "/timeline/entry/event")
+            outputStream << "Error: no such element " << node.path() << "/" << xmlElement << std::endl;
+        } else {
+          if (field->is_repeated()) {
```
JoshDreamland: This block seems to be logic to translate Pugi → Proto. Can you extract it into its own method?

CommandLine/emake/Main.cpp
```diff
@@ -63,14 +69,23 @@ int main(int argc, char* argv[])
       std::string ext;
       size_t dot = input_file.find_last_of('.');
       if (dot != std::string::npos) ext = tolower(input_file.substr(dot + 1));
-      if (ext != "sog") {
+      if (ext == "sog") {
+        if (!ReadSOG(input_file, &game)) return 1;
+      } else if (ext == "gmx") {
+        boost::filesystem::path p = input_file;
```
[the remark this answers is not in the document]
RemoveRusky: all this logic does is translate game.gmx to game.gmx/game.project.gmx so the cli can accept either. I think it's fine here

CommandLine/protos/resources/Background.proto
RemoveRusky: We want to pass the filname not the imgdata so that we can either load the image from path at runtime or pack it at compiletime depending on theb mode

CommandLine/emake/Main.cpp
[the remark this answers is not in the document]
RobertBColton: Please use C++17: https://stackoverflow.com/a/43281413

CommandLine/protos/resources/Font.proto
RemoveRusky: I'm not sure what it is. I've never seen anything other than texture0 in a gmx but I added it because I wanted to cover all the gmx fields

CommandLine/protos/resources/GameInformation.proto
RemoveRusky: pretty sure i copied direct from xml so you can thank yoyo

CommandLine/protos/resources/Sprite.proto
RemoveRusky: "If your sprite is going to be used as a texture map for a 3D game, then you should check the Used for 3D box and the sprite will be given a texture page all of its own. Note: This will increase the texture memory needs of your game tremendously and so great care must be taken when using this option. Also note that all 3D textures must be a power of 2 (ie> 128x128, 256x256, 512x512 etc...)." <- according to gm manual

Makefile
```diff
@@ -1,33 +1,27 @@
...
-all:
-	$(MAKE) -C lodepng/
-	$(MAKE) -C CommandLine/protos/
-	$(MAKE) -C CommandLine/libGMX/
-	$(MAKE) -C CompilerSource/
-	$(MAKE) -C CommandLine/emake/
-	$(MAKE) -C CommandLine/testing/
 
-Game: liblodepng
+all: liblodepng libProtocols libGMX ENIGMA emake test-runner .FORCE
```
JoshDreamland: Let me explain, a little. The `.FORCE` target is a dummy target I created that is never satisfied. It causes `make` to run the rules of a target, even when that target exists, because that target will always appear up-to-date unless you read its actual makefile. So that's what it's forcing to happen.

The `.PHONY` rule is a special rule that tells make that the file doesn't actually exist.

RemoveRusky: whatever, did I use it incorrectly or someshit? if so tell me what to change

RemoveRusky: or better yet, fix it yourself

CommandLine/libGMX/gmx.cpp
RemoveRusky: moving this shit requires a function that takes ~50 variables and makes things harder to read.

CommandLine/emake/Main.cpp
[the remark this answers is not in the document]
RemoveRusky: no. emake already uses boost for other shit and we dont support c++17. This would require upgrading gcc on every travis and would up our build time to ~two weeks

CommandLine/protos/resources/Object.proto
RemoveRusky: Not currently. The gmx parser is set to go up xml elements along with the protobufs. It can go down the ladder in the xml but not up.

RemoveRusky:
@RobertBColton you need to fix the rest kthx

CommandLine/libGMX/gmx.cpp
JoshDreamland: That's fine. Maybe add a comment to each branch (if/else block) explaining what the branch does and why it was taken.

CompilerSource/compiler/components/write_room_data.cpp
```diff
@@ -198,7 +198,7 @@ wto.open((codegen_directory + "Preprocessor_Environment_Editable/IDE_EDIT_roomcr
     {
       wto << "variant room_"<< es->rooms[i].id <<"_instancecreate_" << it->first << "()\n{\n  ";
       if (mode == emode_debug) {
-        wto << "enigma::debug_scope $current_scope(\"'instance creation' for instance '" << it->first << "'\");\n";
+        wto << "enigma::debug_scope $current_scope(\"'instance creation' for instance '" << it->first << "'\");\n  ";
```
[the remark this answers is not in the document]
RobertBColton: Are you sure? What about line 199? That is where I copied the style from, because it seems to be what you were already doing. Unless, you are not the one who wrote line 199.

I went and looked at the source back about 8 years and it seems this has been the way it always was in various places:
https://github.com/enigma-dev/enigma-dev/blame/19c85d9f80eef885338dc0c3395ff1117a702c93/CompilerSource/compiler/components/write_room_data.cpp#L167

I don't think you want me to correct this now because I'd probably have to refactor this whole file and I believe we don't want to mixup those kinds of changes. Do it in a separate pr after this is merged?

CompilerSource/compiler/components/write_object_data.cpp
```diff
@@ -722,7 +722,7 @@ static inline void write_script_implementations(ofstream& wto, EnigmaStruct *es,
     }
     wto << ")\n{\n  ";
```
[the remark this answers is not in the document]
RobertBColton: Again, I'll have to refactor the whole file... see above comment.

CompilerSource/compiler/components/write_globals.cpp
```diff
@@ -62,70 +62,71 @@ int lang_CPP::compile_writeGlobals(EnigmaStruct* es, parsed_object* global)
...
-    //  wto << i->second->type << " " << i->second->prefixes << i->second->name << i->second->suffixes << ";" << endl;
-    wto << endl;
-
-    wto << "namespace enigma" << endl << "{" << endl << "  struct ENIGMA_global_structure: object_locals" << endl << "  {" << endl;
-    for (deciter i = dot_accessed_locals.begin(); i != dot_accessed_locals.end(); i++) // Dots are vars that are accessed as something.varname.
-      wto << "    " << i->second.type << " " << i->second.prefix << i->first << i->second.suffix << ";" << endl;
-
-    wto << "    ENIGMA_global_structure(const int _x, const int _y): object_locals(_x,_y) {}" << endl << "  };" << endl << "  object_basic *ENIGMA_global_instance = new ENIGMA_global_structure(global,global);" << endl << "}";
-    wto << endl;
+  wto << license;
```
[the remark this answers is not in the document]
RobertBColton: I looked over it and I don't think I did. I originally opened it to find out why the CLI was segfaulting at first, and it was just because game info didn't have a string allocated or something. It will be nice when all of this is rewritten with protos instead of having proto2es because then none of this code will crash, because protobuf returns valid defaults.

CommandLine/protos/game.proto
RobertBColton: I don't get what you mean. The fact that it's a domain specific language? The fact that it's easy for a user to add a custom resource type? The fact that it can be changed in backwards-compatible ways? The fact that accessing unset properties doesn't segfault? The fact it doesn't need to look for Java?

CommandLine/libGMX/gmx.cpp
RobertBColton: Well sounds have `<origname>sound\audio\soundname.ext</origname>` and `<data>soundname.ext</data>` whereas sprites and backgrounds have only `<data>images\bkgsprname.png</data>`

So, it should be clear that what I am doing is checking if the end of the full dir is at the beginning of the data/origname fields. Then I know whether to prepend the whole thing or just a part of it.

I was consider replacing `(file_path) = true` custom FieldOption with `(file_path) = "the\prefix\for\this\file\field"`

What do you think about that instead?

CommandLine/libGMX/gmx.cpp
RobertBColton: Are you talking about making them translation unit local so they do not clash with the copy that's also in CompilerSource at linktime? How is not ok with them already being in the GMX namespace? You are afraid of somebody adding `using namespace gmx;` somewhere in CompilerSource (which could very well happen as we move proto2es and deprecated backend/)?

CommandLine/libGMX/gmx.cpp
RobertBColton: Yeah same goes for string_replace_all and some other utility methods. We should maybe take another look at this in a future branch to eliminate some of our code redundancy between the compiler and engine. I believe there's also a lot of duplicate file and directory manipulation code.

CommandLine/libGMX/gmx.cpp
RobertBColton: It would also be possible to make a context struct and pass the context around. I'm not too worried about this, this function is pretty short, and it's also like the only thing to this file. Obviously it's going to need some refactoring to make it double as a JSON deserializer for GMS2 projects. So we can just wait until then to break this thing down farther.

Makefile
RobertBColton: I believe he is saying that `.FORCE` is not needed with a `.PHONY` target because it's already going to run anyway. `clean` does need to be forced because it will never exist either. `all` should also need to be forced for the same reason as `clean`.

CompilerSource/compiler/components/module_write_backgrounds.cpp
```diff
@@ -65,11 +65,10 @@ int lang_CPP::module_write_backgrounds(EnigmaStruct *es, FILE *gameModule)
 
   for (int i = 0; i < back_count; i++)
   {
-      writei(es->backgrounds[i].id,gameModule); //id
+    writei(es->backgrounds[i].id,gameModule); //id
```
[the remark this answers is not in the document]
RobertBColton: Really? I didn't make any hard to decipher changes? Ok though, if you insist.

CompilerSource/Proto2ES.cpp
```diff
@@ -0,0 +1,737 @@
...
+    case ArgumentKind::ARG_BOTH:
+      // treat as literal if starts with quote (")
+      if (val[0] == '"' || val[0] == '\'') return val;
+      // else fall through
+    case ArgumentKind::ARG_STRING:
+      return '\"' + string_replace_all(string_replace_all(val, "\\", "\\\\"), "\"", "\"+'\"'+\"") + '\"';
+    case ArgumentKind::ARG_BOOLEAN:
+      return std::to_string(val != "0");
+    case ArgumentKind::ARG_SPRITE:
+      return arg.has_sprite() ? arg.sprite() : "-1";
```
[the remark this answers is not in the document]
RobertBColton: These are good suggestions, and yeah, I struggled a little bit trying to eliminate this redundancy. Is it possible to give oneof fields a default value in protos? I think I prefer to just put it there, on each resource field, but I'm not sure.

CompilerSource/Proto2ES.cpp
```diff
@@ -0,0 +1,737 @@
...
+  using buffers::resources::ArgumentKind;
+  std::string val = arg.string();
+
+  switch (arg.kind()) {
+    case ArgumentKind::ARG_BOTH:
+      // treat as literal if starts with quote (")
+      if (val[0] == '"' || val[0] == '\'') return val;
+      // else fall through
+    case ArgumentKind::ARG_STRING:
+      return '\"' + string_replace_all(string_replace_all(val, "\\", "\\\\"), "\"", "\"+'\"'+\"") + '\"';
```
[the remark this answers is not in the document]
RobertBColton: I literally copied and pasted it from the compiler. I was afraid to include it probably because I didn't want other CompilerSource code bloating this layer. But I just realized this proto2es layer is compiled into CompilerSource anyway, so ok, I'll just include it.

CompilerSource/compiler/compile.cpp
```diff
@@ -150,10 +150,22 @@ inline void write_exe_info(const std::string codegen_directory, const EnigmaStru
 
 #include "System/builtins.h"
 
-dllexport int compileEGMf(EnigmaStruct *es, const char* exe_filename, int mode) {
+static inline int compileEGMf_Raw(EnigmaStruct *es, const char* exe_filename, int mode) {
```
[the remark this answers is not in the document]
RobertBColton: Ok, I'll remove it.

CommandLine/emake/Main.cpp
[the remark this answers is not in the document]
RobertBColton: Josh has directory functions that can be included from CompilerSource without adding boost into the CompilerSource. I'm also wanting to switch us to deepflags or something else available on pacman. I absolutely hate boost right now because it's not doing enough for us. Unless you decide to start using it for image and sounds too, then I'll reconsider. But for right now, we're simply not getting enough use of it to justify its inclusion.

CommandLine/protos/resources/Object.proto
RobertBColton: Lol maybe we just add support for "../" then? Does XPATH have a standard for that?

CommandLine/protos/resources/Path.proto
RobertBColton: Yes it is, you don't need this in the engine at all. In GMS2 paths are a subresource of rooms I think.

CommandLine/protos/resources/Sprite.proto
RobertBColton: It just means you are going to use it for a 3D model Josh, so it doesn't get added to a texture page so that you can use it with multitexturing. The reason you have to do this for 3D is because texture interpolation will cause other textures on the same texture page to bleed over on your model. However, if you are really good, Rusky says there's a way to do interpolation with paged 3D textures. I asked @rpjohnst about this for iPhone and Android games where computing resources are obviously limited for 3D graphics.

CommandLine/protos/resources/Sprite.proto
RobertBColton: I don't know what you are saying here.

CommandLine/protos/resources/GameInformation.proto
RobertBColton: I did this, and I believe I copied it from the old backend. This is actually deprecated in GMX, only the rtf file itself is included, and YoYo only kept it for making notes, so you cannot in anyway show it at runtime. In GMS2 they created a dedicated Notes resource group which is just text files. So what ordering would you like it to have?

CommandLine/protos/resources/Font.proto
RobertBColton: Well that's what it is, but why? Then we have to add the gmx custom option, which is ok, it makes our format more human-readable. I'll change it.

CommandLine/protos/resources/Font.proto
RobertBColton: I don't know what you are saying.

CommandLine/protos/resources/Background.proto
RobertBColton: I can make it a oneof to support both?

CommandLine/protos/resources/Font.proto
RobertBColton: Yes but @JoshDreamland do not be fooled by the number 0, there is only ever one, so our use of the gmx option here is safe. Textures cannot be added to more than one texture group, at least in GMS 1.4.

CommandLine/libGMX/gmx.cpp
RemoveRusky: this library will be linked against ide and whatever else so it needs its own string methods

RobertBColton: Extract the string functions into a shared library... just like you did for protos.

RemoveRusky: thats overkill for 3 line method

RobertBColton: It's not though, we're also copying string_replace_all around and you're including boost for directory functions Josh already wrote in CompilerSource. Again, extract all this shit, build as shared library.

RemoveRusky: no point. dont need 50 shared libraries. I use boost for dir cause I use it for flags already

RobertBColton:
@RemoveRusky Why exactly do you need EGM_NESTED when all of that room metadata is nested in `<makerSettings>` why can't you just put a FieldOption (gmx) = "makeSettings" on the RoomEditorMetadata field?

RemoveRusky:
I removed all that shit. nested is for the physics shit that josh wanted in a message but isnt in gmx

RobertBColton:
No, I mean, I'm telling you to put (gmx) = "makerSettings" as a MessageOption not a FieldOption. That way you don't have to keep typing EGM_NESTED over and over again for each field. I was also saying to make GMX_SPLIT a MessageOption.

RemoveRusky:
Please learn to read dumbass. I completly removed marker settings fom the game proto EGM NESTED has NOTHING to do with those feilds. It's for the physics feilds you twat

RemoveRusky:
@RobertBColton I refactored this to be tree based but seems I broke proto2es . Could you fix the name2id bits for me? and idk what's going on with the other errors.

RobertBColton:
I am closing this pull request for now so we don't keep eating up our CI slots. I will send a fresh one when work in the branch is completed.

RemoveRusky:
don't close it.

RobertBColton:
What you just committed, it doesn't build on Windows again now. The makefile adds the include directory. If you keep putting that include in the header, then everything that includes that header must also include pugixml, instead of just the GMX reader. Please undo that commit.

RemoveRusky:
Can't it needs to be there. I need to use pugixml in the template function. Your method of seperating template doesnt compile and breaks my shit.

RemoveRusky:
looks like you broke ./emake & ./emake --help

RobertBColton:
Yes, please help me fix that I was trying before I went to bed but otherwise got more hooked up RGM side.

RemoveRusky:
[Army.gmx.gmx.zip](https://github.com/enigma-dev/enigma-dev/files/1858499/Army.gmx.gmx.zip)

RemoveRusky:
You've only hidden the problem. Having it green this way is worse than having it red

RobertBColton:
image

RobertBColton:
And after a few more commits, I have the Wild Racing GMK running. I fixed the temporary file cleanup and a few other issues, including where I made a mistake by reading the stream looking at an individual resources version number instead of the group version number. That was causing me the stream to be misread about whether a resource was deleted. Seems to be working pretty good after all these fixes. The only problem left is that LGM loads 800/810 projects faster, something is really slow about my image reading.

>$ ./emake "C:\Users\Owner\Desktop\GMGames/Wild Racing.gmk" -e "DataStructures,Paths" -o /tmp/test.exe -r

Also, I am unsure if these graphical glitches exist on master or not, but I don't think any of this is related to that.

GMK Wild Racing

RobertBColton:
Alright, I used std::chrono high resolution timing and cut the read time of that Wild Racing game from 8.8 seconds down to around 2.4 seconds. I was measuring in nanoseconds for accuracy as well.

I found from benchmarking that all of the slowdown was in `lodepng_encode32` and discovered that, not only because of debugging symbols, lodepng is the slowest PNG loading library. @RemoveRusky also shared with me the results of someone else's benchmark on PNG loading:
https://caedesnotes.wordpress.com/2015/03/23/comparing-performance-stb_image-vs-libjpeg-turbo-libpng-and-lodepng/

lodepng is significantly slower with and without debugging symbols and optimizations. I've seen some people elsewhere comment that it may be an insecure/unsafe library:
https://news.ycombinator.com/item?id=8600634

We should probably remove lodepng and use libpng, so we can grab it with pacman too and get it out of our codebase, or something else.

RobertBColton:
Ok, I went a little further and optimized it even better by using `std::async` instead of `std::thread` because it's backed by a thread pool, meaning it will amortize the cost of thread creation using some limit (i.e. the number of processor cores). It seems to have shaved off an average 100ms of the total load time too.

RobertBColton:
Some tasks:

- [ ] Fix GMX collisions in AI tutorial
- [ ] Verify thread safety of atexit in GMK
- [ ] Try out libpng instead of lodepng

RobertBColton:
After some more fixes in 035ab378ed76f7f776947cbc2beb51f3e45176d5 I am able to play this AI tutorial gmx project somebody sent me a while back.

AI Tutorial GMX

RemoveRusky:
[gm530.gm6.zip](https://github.com/enigma-dev/enigma-dev/files/1892445/gm530.gm6.zip)

CompilerSource/compiler/components/write_globals.cpp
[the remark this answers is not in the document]
RobertBColton: Please review #1211 so I can get it merged and then merge branch master into this pr resolving your concerns.

CompilerSource/compiler/components/write_object_data.cpp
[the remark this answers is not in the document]
RobertBColton: I've split these changes out into #1211, see above comment.

CompilerSource/compiler/components/write_room_data.cpp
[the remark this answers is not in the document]
RobertBColton: I've split these changes out into #1211, see above comment.

CompilerSource/compiler/components/module_write_backgrounds.cpp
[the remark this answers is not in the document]
RobertBColton: I've split these changes out into #1211, see below comment.

CompilerSource/compiler/compile.cpp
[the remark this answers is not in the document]
RobertBColton: I took care of it in 30c48d305a44626495438967e57de7aef8ae4275. This is actually better because now it means that if this pr were merged, it would not break anybody on the existing portable ZIP setup if they tried to pull. It means this pr does not add any new dependencies for them unless they try to manually build the protos.

RobertBColton:
Nice, in one fell swoop you manujed to totally fuck up my green CI record on this PR. Way to go fucking azzhat.

JoshDreamland:
Merged in #1248.

PR #1184, opened 2018-02-20
fixes some mem leaks related to font crap I'm working on

RemoveRusky:
rect_packer is still duplicated. I'm not sure who should own it. Perhaps we need a a shared folder. Eitherway, atleast one makefile will need modification to remove one of the copies

PR #1185, opened 2018-02-20
font_add

RemoveRusky:
no tests cause Josh lazy


RemoveRusky:
This won't build without #1184

ENIGMAsystem/SHELL/Universal_System/Extensions/ttf/ttf.cpp
```diff
@@ -0,0 +1,187 @@
...
+        pxdata[index] = 255;
+      }
+    }
+    
+    for (const GlyphPair &g : glyphPairs) {
+      error = FT_Load_Char(face, g.index + first, FT_LOAD_RENDER);
+      
+      #ifdef DEBUG_MODE
+      if (error != 0)
+        std::cerr << "Freetype error loading bitmap for char: " <<  static_cast<char>(g.index) << std::endl;
```
JoshDreamland: It might be best to continue the loop, in this case.

PR #1198, opened 2018-03-01
Robert is an idiot

RemoveRusky:
enables undefined behavior runtime checks in test harness and fixes some undefined behavior in game startup

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/fontinit.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/fontinit.cpp
@@ -27,7 +27,8 @@
 #include "Platforms/platforms_mandatory.h"
 #include "Widget_Systems/widgets_mandatory.h"
 
-#include <stdio.h>
+#include <cstring>
+#include <cstdio>
 #include <string>
 
 namespace enigma {
@@ -37,7 +38,7 @@ void exe_loadfonts(FILE* exe) {
   float advance, baseline, origin, gtx, gty, gtx2, gty2;
 
   if (!fread(&nullhere, 4, 1, exe)) return;
-  if (nullhere != *(int*)"FNT ") return;
+  if (memcmp(&nullhere, "FNT ", sizeof(int)) != 0) return;
 
   if (!fread(&fontcount, 4, 1, exe)) return;
   if ((int)fontcount != rawfontcount) {
@@ -84,7 +85,7 @@ void exe_loadfonts(FILE* exe) {
       return;
     }
     if (!fread(&nullhere, 4, 1, exe)) return;
-    if (nullhere != *(int*)"done") {
+    if (memcmp(&nullhere, "done", sizeof(int)) != 0) {
       printf("Unexpected end; eof:%s\n", feof(exe) ? "true" : "false");
       return;
     }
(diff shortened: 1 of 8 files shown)


RobertBColton:
I was just going to send a pr for this as soon as you do.

RemoveRusky:
u slow. I didnt use vectors tho cause cba

RobertBColton:
No, couldn't anyway, Josh said that specific method won't work with a vector, this only way.

ENIGMAsystem/SHELL/Universal_System/fontinit.cpp
```diff
@@ -37,7 +38,7 @@ void exe_loadfonts(FILE* exe) {
   float advance, baseline, origin, gtx, gty, gtx2, gty2;
 
   if (!fread(&nullhere, 4, 1, exe)) return;
-  if (nullhere != *(int*)"FNT ") return;
+  if (memcmp(&nullhere, "FNT ", sizeof(int)) != 0) return;
```
JoshDreamland: This is not as fast... but it is endianness-agnostic, so good change.

ENIGMAsystem/SHELL/Universal_System/image_formats.cpp
```diff
@@ -319,7 +319,8 @@ int image_save_png(string filename, const unsigned char* data, unsigned width, u
     file.write(reinterpret_cast<const char*>(buffer), std::streamsize(buffersize));
     file.close();
   }
-  delete[] buffer;
+  
+  free(buffer);
```
JoshDreamland: If I'm being anal, it might be better to null these pointers out before passing them to libPNG, so they are never left dangling. I don't know if libPNG guarantees that it is safe to `free()` its pointer in an error case.

RemoveRusky: https://github.com/enigma-dev/enigma-dev/blob/master/ENIGMAsystem/SHELL/Universal_System/lodepng.h#L155 It doesn't say to do anything other than free here.

RemoveRusky:
appveyor crapped out on last test for no reason and I dont have access there for some reason

RobertBColton:
AppVeyor does not have the ability to restart a single job, so now it must start completely over...

ENIGMAsystem/SHELL/Universal_System/image_formats.cpp
RobertBColton: @JoshDreamland do not confuse lodepng and libpng, the latter is only a reference implementation.

@RemoveRusky What happens if lodepng has an error reading an image?

RemoveRusky:
I have no access on appveyor at all so i wouldnt know :P

ENIGMAsystem/SHELL/Universal_System/image_formats.cpp
RemoveRusky: the source is right there in our tree you can look yourself

RemoveRusky: anyways do you just want unsigned char* buffer = nullptr; instead of unsigned char* buffer; or what?

JoshDreamland: Yes, I think that's the safest thing. Just in case the LodePNG call returns without nulling the pointer itself.

RemoveRusky:
@RobertBColton's build actually segfaulted on travis but it was still marked green. I added a change to the yaml to prevent that from happening again but I had to shut off the memory leak report on travis until we plug all of enigma's leaks.

JoshDreamland:
I was wondering why nothing bad happened as a consequence of his change... thanks for taking care of that.

RobertBColton:
change title trick

See, I knew he was um big gay.

PR #1204, opened 2018-03-22
widen() and shorten() can now be used everywhere

time-killer-games:
Enables use of these functions so that UTF-8 can potentially be supported on all Windows functions.

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/estring.h
+++ b/ENIGMAsystem/SHELL/Universal_System/estring.h
@@ -26,9 +26,7 @@
 
 #if CURRENT_PLATFORM_ID == OS_WINDOWS
 
-#include <windows.h>
 #include <wchar.h>
-#include <vector>
 
 typedef std::basic_string<WCHAR> tstring;
 tstring widen(const std::string &str);
--- a/ENIGMAsystem/SHELL/Universal_System/estring.h
+++ b/ENIGMAsystem/SHELL/Universal_System/estring.h
@@ -26,6 +26,7 @@
 
 #if CURRENT_PLATFORM_ID == OS_WINDOWS
 
+#include <windows.h>
 #include <wchar.h>
 
 typedef std::basic_string<WCHAR> tstring;
--- a/ENIGMAsystem/SHELL/Universal_System/estring.h
+++ b/ENIGMAsystem/SHELL/Universal_System/estring.h
@@ -22,6 +22,20 @@
 
 #include <string>
 
+#include "CompilerSource/OS_Switchboard.h"
+
+#if CURRENT_PLATFORM_ID == OS_WINDOWS
+
+#include <windows.h>
+#include <wchar.h>
+#include <vector>
+
+typedef std::basic_string<WCHAR> tstring;
+tstring widen(const std::string &str);
+std::string shorten(tstring str);
+
+#endif
+
 namespace enigma_user {
 
 std::string base64_encode(std::string const& str);
--- a/ENIGMAsystem/SHELL/Universal_System/estring.h
+++ b/ENIGMAsystem/SHELL/Universal_System/estring.h
@@ -22,7 +22,7 @@
 
 #include <string>
 
-#include "CompilerSource/OS_Switchboard.h"
+#include "../../../CompilerSource/OS_Switchboard.h"
 
 #if CURRENT_PLATFORM_ID == OS_WINDOWS
 
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
@@ -21,6 +21,8 @@
 #include <sstream>
 #include <algorithm>
 #include <unistd.h>
+#include <windows.h>
+#include <wchar.h>
 #include <vector>
 using std::string;
 using std::vector;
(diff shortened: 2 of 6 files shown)


JoshDreamland:
Can you combine this with your other pull request? Doesn't make sense to introduce an unused grade in a pull request.

time-killer-games:
@JoshDreamland I still don't know how to do that. I can't include the new header because because the online compiler will not be able to find a header that hasn't been merged yet, so it won't pass. After you merge these two pull requests, I could simply add the header #include and you could merge that.

RobertBColton:
@time-killer-games You can make multiple commits to your fork on GitHub, that will all be a part of the same pull request. You are sending this pull request from the `patch-10` branch you created on your fork. 

1. So to add another commit to this pr, just go to your fork:
https://github.com/time-killer-games/enigma-dev
2. Select the `patch-10` branch from the branches pulldown:
image
3. Just make more edits and remember to save/commit. When you enter the new message upon file save github will add another commit to this pr.

RobertBColton:
Ok, that looks good, but you could actually just copy `win32utf8.h` into `WINDOWSmain.h` and then delete `win32utf8.h` (all of which is possible from the GitHub UI but make sure you are on the `patch-10` branch, GitHub really likes to switch you back to master when you click on your project's title) because we really don't need a separate header just for that. And I think Josh wanted to actually use shorten for something somewhere, so it's actually being used. The CI bot builds with `-wall` I believe so it will show warnings about unused functions/templates/etc. Other than that I'd go ahead and merge this if you fix those two things.

I'm closing #1203 since you've merged in those changes here together into a single pr. Nice work on your git skills too btw, it gets easier with time 🍰

time-killer-games:
Ok guys, it should be ready to merge. I decided to move shorten and widen not just in the header but also in the cpp files. This is so that utf-8 can be supported regardless of whether the user is using "Win32" or "None" for their dialog system.

time-killer-games:
@JoshDreamland @RobertBColton Ok guys, it should be ready to merge. I decided to move shorten and widen not just in the header but also in the cpp files. This is so that utf-8 can be supported regardless of whether the user is using "Win32" or "None" for their dialog system.

RobertBColton:
This is perfect, you are getting really good. I didn't think about that, but it could later maybe actually even be moved to universal_system and always built into the engine, because who knows if some user may write some extension that needs to widen or shorten strings.

When Travis and AppVeyor pass, which they probably will, I'll go ahead and merge. 👍

time-killer-games:
@RobertBColton If someone wants to include these functions in their extension, all they have to do is:
#include "Platforms/Win32/WINDOWSmain.h"
That's what I'm currently doing with one of my ENIGMA extensions.

But if you want me to still move it, it's better now, than later when it could potentially break something.

RobertBColton:
That won't work because if the Window system is set to none then none of the .cpp files in the Windows folder are actually compiled. So not only do they have to include the header, they also have to add WINDOWSmain.cpp into their extensions Makefile SOURCES. This actually would break dialogs if they used right now if you switched the Window system to none. So, yeah it probably is better to just go ahead and put it in Universal System now. I'd put it in the sources where all the string functions are.

time-killer-games:
Oh, I gotcha. So how would you like it done? What files under universal_system do you want it in?

time-killer-games:
@RobertBColton what file do you want it in? estrings.h?

RobertBColton:
I'm sorry, I was working on https://github.com/enigma-dev/RadialGM/pull/15/commits/515608ba463f7bd3673d76305ddcfe4392b98074

Yes, estring.h/.cpp is where I was thinking, and no it won't break non-windows, because you should wrap it in:
```cpp
#ifdef _WIN32
#endif
```
We will only support the function on Windows platforms.

time-killer-games:
@RobertBColton it isn't recognizing the path you specified to "CompilerSource/OS_Switchboard.h"


The file does exist, however.

What should I change this line to?:
#include "CompilerSource/OS_Switchboard.h"

RobertBColton:
Prepend the include with `../` to go up the number of directories from `enigma-dev\ENIGMAsystem\SHELL\Universal_System`

I think you'll need: `../../../CompilerSource/OS_Switchboard.h`

Forgive me I'm going to bed for a bit, I've been up trying to tweak some minor issues I have with Qt Creator that I just wanted to see if I could fix and trying to submit a patch to the MinGW packages project for QScintilla. I'll be back up later if you still haven't gotten it by then.

time-killer-games:
@RobertBColton Okay, we should be all set! :) Merge away! (unless I missed something)

ENIGMAsystem/SHELL/Universal_System/estring.h
```diff
@@ -22,6 +22,20 @@
 
 #include <string>
 
+#include "../../../CompilerSource/OS_Switchboard.h"
+
+#if CURRENT_PLATFORM_ID == OS_WINDOWS
+
+#include <windows.h>
+#include <wchar.h>
+#include <vector>
```
JoshDreamland: You aren't using `vector`, here, and this can cause problems in the current function parser. There's technically no sense including `windows.h` for this; we know our implementation supports `wchar_t`, but since this is Windows-only for now, I guess it's fine.

time-killer-games: Ok, I removed windows.h and vector

time-killer-games:
@JoshDreamland Ok dude, we're all set.

JoshDreamland:
Thanks for the contribution!

PR #1246, opened 2018-05-25
Merge #1179 piecemeal

JoshDreamland:
That pull request is massive and should be merged in small, green chunks.


JoshDreamland:
Actually, let's not.

PR #1296, opened 2018-07-02
get_directory/_alt now return dir with final slash

time-killer-games:
to match how GMS works with its directory functions and constants i.e. working_directory

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/TinyFileDialogs/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/TinyFileDialogs/dialogs.cpp
@@ -373,8 +373,8 @@ namespace enigma_user
       char cstr_path[PATH_MAX];
       strncpy(cstr_path, path, PATH_MAX);
 
-      char *slash = (char *)"/";
-      strcat(cstr_path, slash);
+      char slash[] = "/";
+      strncat(cstr_path, slash);
 
       str_result = string_replace_all(cstr_path, "//", "/");
     }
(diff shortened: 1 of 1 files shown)


ENIGMAsystem/SHELL/Widget_Systems/TinyFileDialogs/dialogs.cpp
```diff
@@ -361,7 +361,27 @@ namespace enigma_user
...
+
+    static string str_result;
+    
+    if (ok_pressed == true)
+    {
+      char cstr_path[PATH_MAX];
+      strncpy(cstr_path, path, PATH_MAX);
+
+      char *slash = (char *)"/";
+      strcat(cstr_path, slash);
```
JoshDreamland: Prefer `char slash[] = "/"`, if you need this to be non-const char. Also, this `strcat` does not respect the buffer size; you would need to use `strncat` or add `1` to `PATH_MAX` in the boundary above.

ENIGMAsystem/SHELL/Widget_Systems/TinyFileDialogs/dialogs.cpp
```diff
@@ -361,7 +361,27 @@ namespace enigma_user
...
+    bool ok_pressed = true;
+
+    if (path == NULL)
+      ok_pressed = false;
+
+    static string str_result;
+    
+    if (ok_pressed == true)
+    {
+      char cstr_path[PATH_MAX];
```
JoshDreamland: I think you would benefit from taking a mathematical approach to reasoning about your code. You seem to let Win32-isms build your code up to be more complex than you're comfortable reasoning about. That happens to all of us, but you should go over and think about what elements of the API forced you to code in a certain way. Even without delving too deep into individual API calls, you can reason about the final effects of a block of code and how to achieve the same thing more easily.

In this case, there's no reason to use all these C routines; we're going to copy the final buffer into a `std::string`, anyway, when you pass it to `string_replace_all`. All of the code in this pull request boils down to `return path ? string_replace_all(path + "/", "//", "/") : "";`.

The way it is written contains unnecessary copies, one of which is unsafe.

PR #1302, opened 2018-07-03
more mouse fixes

RemoveRusky:


ENIGMAsystem/SHELL/Platforms/SDL/Window.cpp
```diff
@@ -113,7 +115,11 @@ int handleEvents() { return eventHandler.processEvents(); }
 
 namespace enigma_user {
 
-void io_handle() {}
+void io_handle() {
```
RemoveRusky: this can prolly be used as io_handle for all systems but I'll do that in a separate commit

PR #1353, opened 2018-08-09
Fixed GL1, DX9, DX11 and NONE not compiling with surface_create

HitCoder9768:
Fixed GL1, DX9, DX11 and NONE not compiling with surface_create due to new parameters

Surface create on the newer renderers has
`int surface_create(int width, int height, bool depthbuffer, bool stencilbuffer, bool writeonly)`
However, with older renderers stencilbuffer and writeonly are not present. Quick workaround allows games to be compiled as normal using these renderers.

The following line is now present in the 4 listed renderer's surface cpp files, adding the two bool parameters.
`int surface_create(int width, int height, bool depthbuffer, bool, bool)`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "files.associations": {
-        "xiosbase": "cpp"
-    }
-}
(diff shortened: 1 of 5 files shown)


RemoveRusky:
wow @RobertBColton always breaking everything

PR #1354, opened 2018-08-15
TFD Fixing Single Quotes in Strings (Mac)

time-killer-games:
lol it turned out to be more commits than I thought it would take.

3 words. Shut. Up.

Anyways this fixes a problem on Mac where if the user displayed a dialog with an apostrophe or single quote in the string the character would be ignored and taken out of the string because escaping wasn't done correctly by the original author of TFD. 

It's fixed! Yay! Let's get NAKED!!!!


time-killer-games:
@JoshDreamland did I do the Makefile right and if so please merge. I don't use wildcards anymore in it so I don't know if the mention of wildcard should really be there but idk what to replace it with or if I should just remove it.

time-killer-games:
why is not passing? wow

RemoveRusky:
ci crapped out I restarted job for you

ENIGMAsystem/SHELL/Widget_Systems/TinyFileDialogs/Makefile
```diff
@@ -1,2 +1,9 @@
 include $(SHARED_SRC_DIR)/tinyfiledialogs/Makefile
-SOURCES += $(wildcard Widget_Systems/TinyFileDialogs/*.cpp)
+
+ifeq ($(TARGET-PLATFORM), MacOSX)
+	SOURCES += $(wildcard Widget_Systems/TinyFileDialogs/dialogs-macos.cpp)
```
RemoveRusky: wildcard is for when you have * you can just use SOURCES += Widget_Systems/TinyFileDialogs/dialogs-macos.cpp here

PR #1373, opened 2018-08-30
Fixes bug with get_color() not working on macOS

time-killer-games:


shared/tinyfiledialogs/tinyfiledialogs.cpp
```diff
@@ -1447,18 +1441,9 @@ char const * tinyfd_colorChooser(
...
-                strcat( lDialogString , " -e \"try\" -e \"set mycolor to choose color default color {");
-            }
-            else
-            {
-                strcat( lDialogString ,
-                       " -e \"try\" -e \"tell app (path to frontmost application as Unicode text) \
-                       to set mycolor to choose color default color {");
-            }
-            
+            strcat( lDialogString , " -e \"tell application \\\"System Events\\\"\" -e \"Activate\"");
```
JoshDreamland: This method is so needlessly inefficient, and so error-prone, I'm amazed it is checked in as part of an official codebase somewhere. This whole file is a gas tanker coated in accidents waiting to happen. But I guess you didn't manage to make it any worse, so... proceed...

PR #1393, opened 2018-09-25
Added KDialog Support

time-killer-games:
With the addition of KDailog support, this pr splits the former "Tiny File Dialogs" widget system into the 3 separate widget systems - "OsaScript" (Mac OS X), "Zenity" (GTK+ / Linux), and "KDialog" (KDE / Linux).

Here's a 64-bit Linux executable demo + the GM81 source file: [DialogModule.zip](https://github.com/enigma-dev/enigma-dev/files/2432442/DialogModule.zip)

"OsaScript" (Mac OS X) Preview:

macintosh

"Zenity" (GTK+ / Linux) Preview:

ubuntu

"KDialog" (KDE / Linux) Preview:

kdeneon


JoshDreamland:
These changes to upstream TFD are going to be difficult to maintain, going forward. Can you separate your changes out, so that the latest TFD can be pulled in when new versions become available? It may be better to submit those changes as pull requests to TFD directly.

time-killer-games:
@JoshDreamland any version of TFD that is not this highly modified version of it I have made for ENIGMA will not work. We need to stick with the file I made and maintain it ourselves. The original TFD by the original author is EXTREMELY buggy, which I had to patch up from head to toe, all the way down to how individual strings are quoted and escaped. Tr

PR #1414, opened 2018-10-17
Dialog Module Updates

time-killer-games:
Please do not prematurely close this pr. I am going to attempt to make my other commits as a part of this same pr this thyme.

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/DlgMod/Info/About.ey
+++ b/ENIGMAsystem/SHELL/Widget_Systems/DlgMod/Info/About.ey
@@ -3,7 +3,9 @@
 
 Name: DlgMod
 Identifier: DlgMod
-Build-Platforms: Windows
+Build-Platforms: Windows, SDL
 Represents:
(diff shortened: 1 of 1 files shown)


RobertBColton:
Heads up, for RadialGM, I was considering adding a website field for systems and extensions. I am adding such a setting for the Project Info to put the game's website, so I thought it might be a good idea. The link would be clickable and readonly.

RadialGM API Settings
RadialGM Project Info Settings

time-killer-games:
@RobertBColton I'd still like to have a link in the description as Well, so they know what the link is for in this case. Some of them have necessary to know install instructions and important license information, and without knowing this stuff is at the link, people will just try the widgets as they are and not realize why it isn't working out of the box.

RobertBColton:
Right, that's a good point, but also to keep things simple, I think we'd prefer that systems/extensions in the master repo all have GPL or GPL compatible licenses. Regardless, it's still worth considering later because anybody can make and distribute third party systems/extensions/etc.

time-killer-games:
@RobertBColton at the very least it needs to indicate somehow there is install instructions at the link. Zenity and KDialog need an additional terminal command to install, and DlgMod needs an external DLL. All of this is very important to know and we already have had several users complain about Zenity and KDialog not working because they didn't understand they have their own individual dependencies apart from ENIGMA's that need installing.

PR #1448, opened 2018-12-08
Migrate compiler to a GameData structure.

JoshDreamland:
We're currently taking input from two sources: LateralGM (via the EnigmaStruct shared-memory model) and RadialGM (via the new Project) proto.

Passing either of these structures around in the compiler is error; the former is tailored to C-like languages which structure everything in memory manually, and the latter is tailored to the IDE tree and is unfit for quick processing.

By migrating both to a flat model that is proto-based, we can save code and do the flattening up front.

This request is not ready to be merged and is mostly for running incremental builds on the changes.


RobertBColton:
Writing it down this time so you don't forget what I can see is broken:

### The game icon is using the game info icon and I can't close the game.
Wrong Game Icon
### Fonts are invisible or something so no text is drawn in any game
Invisible Fonts
### Backgrounds are missing in some games
Game of Life Black Background
### Room tiles don't appear but `tile_add` seems to work
No Room Tiles

RobertBColton:
Ok, writing down some more things I run into as I work on this.

### Extensions/Constants Missing Proto
Extensions don't have a proto just because I think I forgot to get around to them. I wasn't sure if we were planning to keep supporting GM extensions **and** ENIGMA extensions. There's also another problem, GMSv1.4 had extensions as resources where you could add includes using the tree. As I recall, the stopped supporting "installed" extensions that you place into the GM installation folder. LGM coming from such a time in GM's history only ever had the name and path of an extension but not all of its other metadata. I didn't know what to do, if we're going to do extensions as a resource, installed extensions, or only ENIGMA extensions. One idea I have is we could do the new GM way and support legacy "installed" extensions by pretending they are ENIGMA extensions. Really need your input on this.

There's no constants proto because they are deprecated in GMS2 and I want to deprecate them in ENIGMA too. In GMS2 you can create a script and use the `#macro` directive to create a global constant. I would like us to add this to ENIGMA, since people are requesting it, and then libEGM will generate a script with a `#macro` for each constants resource when loading an old GM project.

RobertBColton:
@JoshDreamland alright: 4553ae24be651573d67a727c3cf381cfa700b374
I imported the game settings from EnigmaStruct and it seems to be working now. I have a game icon, title, and buttons as well as the ability to close the game. The said, I need to clarify to you that we are planning on multiple game settings in the proto (hence why we just have a tree) and that the one that gets written is supposed to be the one in the "selected configuration" combo box.

Also, you seemed to have missed the part about the shortcuts. Me and fundies want to make the game settings allow you to remap the shortcuts to any input device. You would also be able to add new shortcuts and remap them as you wish. By default, we would use entries like "close_game" to determine what input closes the game.

### Game Settings Imported
Game Settings Imported to Game Data

### Game Settings Input Mapping
Game Settings Input Map

### Current Config Selector
Current Config Selector

CommandLine/libEGM/gmk.cpp
```diff
@@ -455,22 +455,23 @@ int LoadSettings(Decoder &dec, Settings& set) {
...
-  gfx->set_window_scale(dec.read4());
+  gfx->set_view_scale(dec.read4());
   if (ver == 530) {
     dec.skip(8); // "fullscreen scale" & "only scale w/ hardware support"
   } else {
-    gfx->set_window_sizeable(dec.readBool());
-    gfx->set_window_stayontop(dec.readBool());
-    gen->set_color_outside_room_region(dec.read4());
+    win->set_is_sizeable(dec.readBool());
+    win->set_stay_on_top(dec.readBool());
```
RobertBColton: @JoshDreamland why did you rename this? I called it `window_stayontop` because the function is `window_set_stayontop` and I thought you wanted to match GML?
http://gamemaker.info/en/manual/405_10_window

JoshDreamland: It reads sort of funny without the `window_` prefix, and we've already broken consistency all over the place for no real reason, but I'm not opposed to putting it back

RobertBColton: Your call boss, I just wanted to point it out 😛

JoshDreamland: I think the underscores won't kill us. In particular, I think it's more intuitive to expect this grammar change, but if not, we can talk about helpful compatibility changes, later.

RobertBColton:
Picking up where we last left off, I made a test project with a background and noticed this branch and master are always writing it with 426619 size. So that rules out that possibility.

I then tested `draw_background` and it actually worked. So the whole tiles/backgrounds issue must be confined to rooms only.

JoshDreamland:
Good find. I remember making that change consciously... I saw some commented fields that were wishful thinking on my part because LGM was so hard to coordinate additions to. I modified the code to actually use proto data now that it's available. I believe I assumed that the fields would have default values. If that's not the case, it should be. The default tile color should be white, and the default tile alpha should be opaque.

If you're avoiding that for some reason, please set those values in the appropriate GameData ctors. You may have noticed that this class functions essentially like ES2Proto.

RobertBColton:
Ok, so the thing here is that we hadn't exactly decided on where defaults should go. Back when I asked you about setting the defaults in the proto, you sort of scoffed at that in favor of "smarter" defaults.  That's why we aren't setting any basic defaults (such as default tile size of 32x32) anywhere in any of the protos. So are you saying it's ok to do that now? Can I pick use the defaults LGM/GM have had for some time?

JoshDreamland:
I'm glad you're keeping that in mind. Let me reiterate; it may make more sense, at this point.

A default can go in the proto **if it's not specific to any one type of game or IDE.** So it's fine to default the tile color to opaque white, because that's a normal tile—the default default of zero will break things while white/opaque will make things work as expected. An example of a bad default is a room color of cornflower blue, because it fits well with the RadialGM color scheme. You can justify cornflower in other ways—eg, it makes black drawings and white drawings both show up by default. The idea is, a default should serve some greater purpose than whimsically not being zero.

RobertBColton:
Ok, so it wasn't just the color and alpha, it was also the scaling. So, anyway, the room background works in Game of Life now and the room tiles work in the Tile Collision Demo again. Those defaults are good enough, I'll probably do a run over the other protos later in another pull request so when the IDE creates new resources they have reasonable defaults.

RobertBColton:
Alright: 2d845ab631feee1e590ffd873175f49670832dc6

I found a typo you made which caused the player to spawn into the floor. With that fixed, we've come full circle and the Tile Collision Demo works great again.
Tile Collision Demo Fixed

RobertBColton:
Ok, so I caught another regression besides the fonts issue that is left. It occurs in the AI Tutorial GMX project I have which builds fine on master. It seems to be related to the object access stuff you changed.

```cpp
C:/Users/Owner/AppData/Local/ENIGMA/Preprocessor_Environment_Editable/IDE_EDIT_objectaccess.h: In function 'var& enigma::varaccess_i(int)':
C:/Users/Owner/AppData/Local/ENIGMA/Preprocessor_Environment_Editable/IDE_EDIT_objectaccess.h:117:7: error: duplicate case value
       case obj_unit_par: return ((OBJ_obj_unit_par*)inst)->i;
       ^~~~
C:/Users/Owner/AppData/Local/ENIGMA/Preprocessor_Environment_Editable/IDE_EDIT_objectaccess.h:116:7: note: previously used here
       case obj_unit_par: return ((OBJ_obj_unit_par*)inst)->i;
       ^~~~
```

RobertBColton:
Ok: 8e2f1149b6da3f55e2543d8970909b19aa55bc11

I've fixed another regression I caught earlier. I noticed the start screen of Wild Racing was red when the default window color is 0x000000FF (opaque black). This was just a mistake in the part where I map the settings and I needed to map the color, which was red because the endianess was reversed, using the `javaColor` helper.

JoshDreamland:
I wasn't going to take the liberty of deleting your code for you. If you'd like to include that deletion in this branch, go for it.

CompilerSource/backend/GameData.cpp
```diff
@@ -0,0 +1,411 @@
...
+    glyphs.emplace_back(range.glyphs[i]);
+  }
+}
+
+FontData::GlyphData::GlyphData(const ::Glyph &glyph):
+    ImageData(glyph.width, glyph.height, glyph.data,
+              glyph.width * glyph.height) { // Glyph images are alpha-only.
+  metrics.set_origin(glyph.origin);
+  metrics.set_baseline(glyph.baseline);
+  metrics.set_advance(glyph.advance);
```
gfundies: need   
'''metrics.set_width(glyph.width);
 metrics.set_height(glyph.height); ''' to fix fonts

gfundies:
> Ok, so I caught another regression besides the fonts issue that is left. It occurs in the AI Tutorial GMX project I have which builds fine on master. It seems to be related to the object access stuff you changed.
> 
> ```c++
> C:/Users/Owner/AppData/Local/ENIGMA/Preprocessor_Environment_Editable/IDE_EDIT_objectaccess.h: In function 'var& enigma::varaccess_i(int)':
> C:/Users/Owner/AppData/Local/ENIGMA/Preprocessor_Environment_Editable/IDE_EDIT_objectaccess.h:117:7: error: duplicate case value
>        case obj_unit_par: return ((OBJ_obj_unit_par*)inst)->i;
>        ^~~~
> C:/Users/Owner/AppData/Local/ENIGMA/Preprocessor_Environment_Editable/IDE_EDIT_objectaccess.h:116:7: note: previously used here
>        case obj_unit_par: return ((OBJ_obj_unit_par*)inst)->i;
>        ^~~~
> ```
> 
> AI Tutorial Object Access GML

how bout you actually supply the broken game. how dafux anyone gonna know what "ai tutorial is" I was going to fix this now but again you cant post all the pieces to the problem and are sleeping

CompilerSource/backend/GameData.cpp
RobertBColton: @gfundies The correct symbol is the [backtick key](https://en.wikipedia.org/wiki/Grave_accent#As_surrogate_of_apostrophe_or_(opening)_single_quote), not a single quote, for code blocks in both Git Markdown and Discord.

CompilerSource/compiler/components/write_object_access.cpp
```diff
@@ -104,26 +104,23 @@ int lang_CPP::compile_writeObjAccess(map<int,parsed_object*> &parsed_objects, pa
...
+    for (parsed_object *parent : parsed_objects) {
       while (parent) {
         map<string,dectrip>::iterator x = parent->locals.find(pmember);
         if (x != parent->locals.end())
         {
           string tot = x->second.type != "" ? x->second.type : "var";
           if (tot == dait->second.type and x->second.prefix == dait->second.prefix and x->second.suffix == dait->second.suffix)
           {
-            wto << "      case " << it->second->name << ": return ((OBJ_" << it->second->name << "*)inst)->" << pmember << ";" << endl;
+            wto << "      case " << parent->name << ": return ((OBJ_" << parent->name << "*)inst)->" << pmember << ";" << endl;
```
gfundies: this is cause of bug @RobertBColton crying about. It appears when an object is used as a parent more than once. I'm not sure what the fix is

JoshDreamland: Right you are. That was a hasty change on my part. Fixed.

JoshDreamland:
Any lingering testing you have to do, please get it out of your system. The fundies is growing impatient. It must be fed.

RobertBColton:
Alright, I tested all of the games in my possession and only ran onto one additional regression. The game information has stopped working on this branch. We've been having discussions of deprecating it as a resource. I would like to add the "Notes" resources from GMS2. We've also talked about keeping game info but just letting you create unlimited ones for creating worklogs in the IDE. Maybe we could continue writing the first one in the tree for the engine for now though. Just letting you know, that's the only thing not working now in this pull request.

JoshDreamland:
My understanding is that you made the conscious decision not to support game information when you wrote the GameSettings copy logic. If you're having second thoughts, feel free to write it now. Or do it later. I think expecting us to ship a rich text viewer with a game engine is silly, regardless, so I won't miss it one way or another.

RobertBColton:
The test harness has two jobs that test loading projects that have timelines. One is a GMK and one is GMX.

RobertBColton:
@JoshDreamland ok, I did some more testing locally and emake seems to be ok. I did notice a change however. Projects that use fonts, but don't have any font resources, now crash on run from emake. Projects that use fonts, but have font resources, do not. This is an obvious change in behavior because on master, the fonts just won't show but in neither case do they crash. Again, this is only in regards to emake, LGM is fine. Theoretically it seems you could have possibly busted some logic that has to do with writing a default font if the user doesn't have any font resources (which is supposed to be done even if they do, in case they never set the font).

RobertBColton:
@JoshDreamland alright, I had to make a specific test for timelines because I don't have any games, other than the CI tests we made too, that actually use them.

Download Timeline Test Projects: [timeline-tests.zip](https://github.com/enigma-dev/enigma-dev/files/2770634/timeline-tests.zip)

PR #1459, opened 2018-12-15
Add GameMaker-compatible joystick ID's...

time-killer-games:
As the GM 8.1 AND GMStudio manual clearly explains, the id starts at 1, NOT zero. Ends at 2.

The second argument of joystick_check_button() also starts at 1. Ends at 32.

GM 8.1:
http://gamemaker.info/en/manual/404_03_joystick

GMStudio:
https://docs.yoyogames.com/source/dadiospice/002_reference/mouse,%20keyboard%20and%20other%20controls/joystick%20input/index.html

No wonder why my joystick was recognized in GameMaker but NOT ENIGMA...


RobertBColton:
I went and checked all of the manuals back to GM5 and you are correct the joystick ids have traditionally been either 1 or 2. I then went and looked at the Windows joystick API that GM and ENIGMA use to discover that the first joystick id has a constant `JOYSTICKID1` which resolves to 0. Considering that, I can't seem to come up with a good explanation for why GM arbitrarily started the ids at 1 instead of 0 for joysticks.

GM5 Manual

time-killer-games:
I don't give a fuck about this useless pr. Poopy doo doo my fucking bullshit.

PR #1498, opened 2019-01-11
[TFD] stop doubles from going past 15 digits

time-killer-games:
I knew about this for quite some time but didn't feel like fixing it until just now, because people technically could fix this issue in their EDL/GML if they really wanted to. This auto-fixes an issue where get_integer() and get_passcode() would would have glitchy default or return value for the textbox if more than 999,999,999,999,999 used as default or returned.

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
@@ -47,7 +47,7 @@ using enigma::tfd_DialogEngine;
 using std::string;
 
 static inline string remove_trailing_zeros(string strnumb) {
-  if (strnumb.find('.') != std::string::npos) {
+  if (strnumb.find('.') != string::npos) {
     while (!strnumb.empty() && (strnumb.back() == '0' || strnumb.back() == '.'))
       strnumb.pop_back();
   }
--- a/ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
@@ -295,7 +295,7 @@ double get_integer(string str, double def) {
   else if (caption == "" && tfd_DialogEngine() == tfd_KDialog)
     caption = "KDialog";
 
-  if (def > std::numeric_limits<double>::max()) 
+  if (def > std::numeric_limits<double>::max())
     def = std::numeric_limits<double>::max();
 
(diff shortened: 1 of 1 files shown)


time-killer-games:
@JoshDreamland this one is ready for you to review.

ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
```diff
@@ -287,9 +294,10 @@ double get_integer(string str, double def) {
   else if (caption == "" && tfd_DialogEngine() == tfd_KDialog)
     caption = "KDialog";
 
-  std::ostringstream def_integer;
-  def_integer << def;
-  string integer = def_integer.str();
+  if (def > 999999999999999) 
+    def = 999999999999999;
```
JoshDreamland: What is the significance of this constant? It's too large to fit into an integer, so the compiler is probably treating it as a long. It's smaller than the maximum storage of a long. Based on the number of bits used, this is the maximum number of nines you can fit in a double (without loss of precision), but that's not meaningful in and of itself—nothing is special about the number nine. The biggest number you can store without loss of precision is `2 << 56 - 1`, and there should be a constant for that, but if your solution is to cap it, why bother? That's just a greater loss of precision.

time-killer-games: @JoshDreamland I wasn't 100% following what you were trying to say. Let me know if my latest commit does what you wanted me to do.

time-killer-games:
Trying to go beyond the size of a double results in a value of random characters rather than an actual number.

JoshDreamland:
Random characters cannot be stored in a double... can you show me an example of this happening?

time-killer-games:
> Random characters cannot be stored in a double... can you show me an example of this happening?

It happens in all my Widget systems prior to the changes this pull request makes:

`get_integer("i told you so:", 9999999999999999999999999999999999999);`

will give you for example:

JoshDreamland:
Those aren't random characters, Sam; that's called scientific notation. It's used when numbers are too large to express normally. In that case, the full result is sixteen digits, so it is represented as 6.874e16 rather than 68740000000000000.

time-killer-games:
Yes but what I'm doing is mimicking what YoYo does; they limit it so that scientific notation doesn't show. If you don't want to adopt this behavior i understand; it's up to you.

time-killer-games:
@JoshDreamland they limit it yes, I never figured out the exact number that is the limit. Once you hit past a certain number it will always retain the same big number lel which honestly i don't even remember what that number is it's been a few months since I last tested this.

PR #1510, opened 2019-01-21
Revert "Migrate compiler to a GameData structure."

JoshDreamland:
Reverts enigma-dev/enigma-dev#1448

The question is, does this fix Travis? If so, we're going through with the rollback. The branch has caused a couple of other regressions, one of which we can now reproduce. The other regression remains unknown, as the user reporting the issue could only scream "OH MY GOD, WHAT IS THIS, I'VE NEVER SEEN IT BEFORE" before ostensibly frothing at the mouth and keeling over.


JoshDreamland:
The answer is, no, this doesn't fix Travis. While rolling this change back would fix a couple of regressions, we're stuck in the muck either way.

time-killer-games:
lol ok. I get paranoid after being talked about behind my back by apparently a lot of people.

PR #1525, opened 2019-01-31
Make Mac dialogs use *.icns file set by Info.plist

time-killer-games:
Before this pull request, the dialogs that display an icon would show the default icon for app bundles that have no icon set. This change makes sure the dialog recognizes and uses the icon set for the app bundle by the Info.plist if the file happens to exist, otherwise fall back on the previous behavior.

Applies to all of the following functions, except show_error() due to the limitations of applescript.

show_message():

show_question():

show_error():

get_string():

get_password():

get_integer():

get_passcode():

Unlike the rest of the above functions, show_error() uses a stop sign icon, which in some versions of OS X, specifically the newer ones will show that default "no icon set" icon in the corner regardless, because applescript does not allow the user to change that icon when the stop sign is used. I'm keeping the use of the stop sign to make it easy to grasp visually without having to read the dialog that it is an error message.

In addition, KDialog now supports multiple file filters! KDialog is also now able to set the default directory and file name for the file and folder dialogs. The only bug remaining in KDialog is not being able to set the default text with get_password() and get_passcode(), (textbox is always initially empty).

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
@@ -144,6 +144,17 @@ static inline string detect_all_files_filter(string filter) {
   return filter;
 }
 
+/* I know you guys don't like using platform macros; I'm ok with alternative ways to do this
+just let me know what you would rather me do and I'll change it accordingly. Thank you!!! */
+static inline void window_activate() {
+  #ifdef __APPLE__
+  if (tfd_DialogEngine == tfd_OsaScript) {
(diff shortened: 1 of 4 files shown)


ENIGMAsystem/SHELL/Widget_Systems/General/tinyfiledialogs/dialogs.cpp
```diff
@@ -260,6 +273,8 @@ int show_message(const string &str) {
   msg = tfd_add_escaping(msg);
   caption = tfd_add_escaping(caption);
 
+  window_activate();
```
JoshDreamland: As the number of repeated lines continues to stack up, you might consider extracting an initialize helper of some sort.

PR #1529, opened 2019-02-03
Made TFD easier to read

time-killer-games:
previously the inconsistent indentation used in TFD was one big mess and made it a tiny bit difficult to read and looked very sloppy. This should fix it for the most part, if not completely.

I kinda wanted to go even further and make this file match the rest of ENIGMA's intended indentation and syntax practices, but I was too afraid I'd accidentally mess up the file somewhere. Maybe some day...

I also promised Josh I would add a helper function in a different file so I did that as well.


time-killer-games:
@JoshDreamland alright I ended up doing some last minute changes but anyway everything is passing again and it is now ready for you to merge. Tested locally and verified working.

PR #1551, opened 2019-02-18
osx clang fix

time-killer-games:
more info here: https://stackoverflow.com/questions/49577343/filesystem-with-c17-doesnt-work-on-my-mac-os-x-high-sierra/49600206


JoshDreamland:
That thread is a year old. Does OS X still not move C++17 libraries out of the "experimental" folder?

time-killer-games:
Adding experimental was the only way I got female building, so yes.

PR #1583, opened 2019-03-10
Add shellscript_evaluate()

time-killer-games:
simple EDL function i wrote i.e. a better alternative to execute_shell()

Windows will be a different pr because I still need to add UTF-8 support and I'm about to go to church.

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
@@ -25,6 +25,7 @@
 #include <limits.h>
 #include "CocoaMain.h"
 #include "ObjectiveC.h"
+#include <string.h>
 #include <cstdlib>
 #include <string>
 
@@ -98,4 +99,19 @@ string parameter_string(int n) {
   return string("");
 }
 
+string shellscript_evaluate(string command, size_t buffer_size) {
+  char buffer[buffer_size];
+  FILE *file = popen(command.c_str(), "r");
+  while (fgets(buffer, buffer_size, file) != NULL) {}
+  pclose(file);
+
+  if (buffer[strlen(buffer) - 1] == '\n')
+    buffer[strlen(buffer) - 1] = '\0';
+
+  char *result = (char *)malloc(strlen(buffer) + 1);
+  strcpy(result, buffer);
+
+  return string(result);
 }
+
+} // namespace enigma_user
--- a/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
@@ -20,7 +20,7 @@
  */
 
 #include <unistd.h>
-#include <stdio.h>
+#include <cstdio>
 #include <mach-o/dyld.h>
 #include <limits.h>
 #include "CocoaMain.h"
--- a/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
@@ -25,7 +25,7 @@
 #include <limits.h>
 #include "CocoaMain.h"
 #include "ObjectiveC.h"
-#include <string.h>
+#include <cstring>
 #include <cstdlib>
 #include <string>
 
--- a/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
@@ -108,10 +108,7 @@ string shellscript_evaluate(string command, size_t buffer_size) {
   if (buffer[strlen(buffer) - 1] == '\n')
     buffer[strlen(buffer) - 1] = '\0';
 
-  char *result = (char *)malloc(strlen(buffer) + 1);
-  strcpy(result, buffer);
-
-  return string(result);
+  return buffer;
 }
 
 } // namespace enigma_user
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
@@ -53,10 +53,7 @@ string shellscript_evaluate(string command, size_t buffer_size) {
   if (buffer[strlen(buffer) - 1] == '\n')
     buffer[strlen(buffer) - 1] = '\0';
 
-  char *result = (char *)malloc(strlen(buffer) + 1);
-  strcpy(result, buffer);
-
-  return string(result);
+  return buffer;
 }
 
 } // namespace enigma_user
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
@@ -3,6 +3,8 @@
 #include <limits.h>
 #include <unistd.h>
 #include <libgen.h>
+#include <string.h>
+#include <stdio.h>
 #include <cstdlib>
 #include <string>
 
@@ -39,6 +41,25 @@ void initialize_directory_globals() {
 }
 
 }
+
+namespace enigma_user {
+
+string shellscript_evaluate(string command, size_t buffer_size) {
+  char buffer[buffer_size];
+  FILE *file = popen(command.c_str(), "r");
+  while (fgets(buffer, buffer_size, file) != NULL) {}
+  pclose(file);
+
+  if (buffer[strlen(buffer) - 1] == '\n')
+    buffer[strlen(buffer) - 1] = '\0';
+
+  char *result = (char *)malloc(strlen(buffer) + 1);
+  strcpy(result, buffer);
+
+  return string(result);
+}
+
+} // namespace enigma_user
 
 int main(int argc, char** argv) {
   return enigma::enigma_main(argc, argv);
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
@@ -3,8 +3,8 @@
 #include <limits.h>
 #include <unistd.h>
 #include <libgen.h>
-#include <string.h>
-#include <stdio.h>
+#include <cstring>
+#include <cstdio>
 #include <cstdlib>
 #include <string>
 


ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
```diff
@@ -25,6 +25,7 @@
 #include <limits.h>
 #include "CocoaMain.h"
 #include "ObjectiveC.h"
+#include <string.h>
```
JoshDreamland: In C++ sources, use `<cstring>` instead.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
```diff
@@ -98,4 +99,19 @@ string parameter_string(int n) {
...
+string shellscript_evaluate(string command, size_t buffer_size) {
+  char buffer[buffer_size];
+  FILE *file = popen(command.c_str(), "r");
+  while (fgets(buffer, buffer_size, file) != NULL) {}
+  pclose(file);
+
+  if (buffer[strlen(buffer) - 1] == '\n')
+    buffer[strlen(buffer) - 1] = '\0';
+
+  char *result = (char *)malloc(strlen(buffer) + 1);
```
JoshDreamland: This is a memory leak.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
```diff
@@ -98,4 +99,19 @@ string parameter_string(int n) {
   return string("");
 }
 
+string shellscript_evaluate(string command, size_t buffer_size) {
+  char buffer[buffer_size];
```
JoshDreamland: This is non-standard (and has been causing us lots of trouble on non-GNU for that reason) and will break on big buffers. Use a char vector or std:: string.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
```diff
@@ -98,4 +99,19 @@ string parameter_string(int n) {
   return string("");
 }
 
+string shellscript_evaluate(string command, size_t buffer_size) {
```
JoshDreamland: Apple is POSIX-compliant and should be able to use this function from a more general POSIX source.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
```diff
@@ -40,6 +42,25 @@ void initialize_directory_globals() {
 
 }
 
+namespace enigma_user {
+
+string shellscript_evaluate(string command, size_t buffer_size) {
+  char buffer[buffer_size];
+  FILE *file = popen(command.c_str(), "r");
+  while (fgets(buffer, buffer_size, file) != NULL) {}
```
JoshDreamland: Have you tested this code? It shouldn't work for multiple lines of text, anyway.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
```diff
@@ -40,6 +42,25 @@ void initialize_directory_globals() {
 
 }
 
+namespace enigma_user {
+
+string shellscript_evaluate(string command, size_t buffer_size) {
```
JoshDreamland: Passing in a `buffer_size` is an unnecessary implementation detail. This function doesn't need one to work (and work correctly).

Also, the return type does not give the user a way to check whether the call succeeded. Usually, if a user cares about data that was printed, they care about whether the routine succeeded as well.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
```diff
@@ -3,6 +3,8 @@
 #include <limits.h>
 #include <unistd.h>
 #include <libgen.h>
+#include <string.h>
+#include <stdio.h>
```
JoshDreamland: Same for `cstdio`.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: I was going to ask where you want the function but you stopped responding to me on hangouts so I just assumed make a pr and hope you'd respond after that

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: How do you suggest I do this differently? Last I checked I need a buffer. Letting the user choose the size is just sort of necessary for preventing a stack overflow. It can also be the reason to cause one, but not if the user knows what they are doing. If I don't let them choose I would just assume a large value that will fit their output no matter what = not a good idea. Only the user knows how many characters they need for output, so they should know to use a smallest size needed.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
time-killer-games: Well, it technically works for multiline output. Whether it does for the command argument I haven't tested, but that could be solved by doing a std::replace() all the new line characters (\n and \r individually) with a whitespace. My code I have tested with has no new lines but it does have stuff that would be easier to read if it had multiple lines such as if statements.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
time-killer-games: Again, how would I go about doing that? I have no idea.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
time-killer-games: Oh, right, the C++ headers are preferred. My bad.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: You didn't tell me how to make it not leak memory. This could save me hours of Google searching.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: I honestly didn't know that there was a c++ version of that header. Thanks for the info.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
JoshDreamland: Typically, the buffer is a fixed width (often `BUFSIZ`) and you read that many bytes into it at a time until the input is exhausted. In this case, you would append each chunk to a string, as it came in (`fread` returns the number of elements read).

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
JoshDreamland: In general, don't use `malloc` (or even `new`) in C++. The number of times you need to call either function in a modern C++ application is essentially zero. And I didn't think I'd need to point out an alternative, because this buffer is completely unnecessary the way this code is implemented (this should have just read `return buffer;`) and unusable the way this code should be implemented: `strcpy` stops at the first null character, so your code, as written, *depends* on the string being null-terminated (which, fortunately, it is). But to support multi-line output of arbitrary size, you'll need to append multiple chunks to the output string, anyway.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: This says your method sucks rooster cock https://stackoverflow.com/questions/51627159/how-to-get-text-using-fgets-and-save-stdstring

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
JoshDreamland: It will not work for multi-line output, as implemented. You always write to the same buffer, right at the beginning. Since `fgets` null-terminates, it will "work" in the sense that the return value will always be the last line printed. Unless the command prints two newlines at the end instead of one, in which case, your code will return the empty string.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
JoshDreamland: Explained above.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
JoshDreamland: Right next to `execute_shell` is probably the best place...

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
time-killer-games: well i can verify from my testing that the multi-line output 100% works, and i have seen quite a few new lines returned

time-killer-games: for instance, here's what selecting multiple files from an zenity open file dialog returns, and surprise-surprise it actually output the new lines:

Screenshot from 2019-03-10 12-01-12

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: sorry my last comment posted here was before my page refreshed and made your new reply visible

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: I just tried returning buffer which was how my code originally was before posted here and yeah
warning: address of local variable ‘buffer’ returned [-Wreturn-local-addr]

time-killer-games: oh i think i know why it's doing that with my code im testing it returns a char * not a std::string. silly me

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: so are you saying i do for instance

buffer[BUFSIZ];
string str_result = "";
...
size_t bytes_read = 0;
while (bytes_read < buffer) {
  bytes_read = fread(buffer, sizeof(buffer), 1, file);
  str_result += buffer;
}

I feel completely lost tbh

JoshDreamland: No, that post says what the individual who asked that was doing sucks rooster cock. The `fgets` routine does not belong here.

time-killer-games: No idea what I use if I can't use that. I guess I Google it since you probably feel like I'm manipulating you into writing code for me and I don't want to come across that way so I'll see what I can do to get this much figured out on my own.

time-killer-games: Aha! std::getline. See? I'm not a complete idiot after all.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
JoshDreamland: The default separator in Zenity is the space character. It would be an operating system bug if your code worked the way you had written it.

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
JoshDreamland: That is a fine approach. You can also just read the raw characters; there's nothing inherently special about newlines.

ENIGMAsystem/SHELL/Platforms/General/POSIX/main.cpp
time-killer-games: I'm not using the default separator. I changed it to \r

JoshDreamland: (Which means there are no newlines in the output.)

time-killer-games: Oh no, but there are. You didn't see the shell script I was executing it appears, even though I sent you it like 5 times on hangouts

time-killer-games: I also know this is not an operating system bug because the TFD source uses fgets too many people have downloaded that thing and no one has complained about multiline output not working once

JoshDreamland: Sam, you might as well argue with me that 2 + 2 is 17. It's impossible for your original implementation to read in a newline character. Each newline encountered, `fgets` will stop. You will call `fgets` again, and it will overwrite the previous line in the buffer. That is the end of the story.

time-killer-games: Ok fine, you win big daddy. But I still say 2 + 2 = 17

ENIGMAsystem/SHELL/Platforms/Cocoa/CocoaMain.cpp
time-killer-games: I guess I'll do it the way Google told me since the way you just said wasn't detailed enough for to even know how how to look that up

time-killer-games: according to the second answer on this: https://stackoverflow.com/questions/43182254/c-equivalent-of-fgets-for-reading-from-popen

"The typical approach here is to read in chunks of, say, 512 bytes. Then, iff you need the entire result in a single buffer, append into a std::string or std::vector<char> inside that loop. Yes, it'll result in a bit of data copying, but your computer can handle it. And you'd always have had to do this anyway."

I have no idea how to write that because it was just as nonspecific as you were. What will the while loop look like? What will cause the while loop to keep going and stop once the read is finnished? What function do i put in the while loop? while (what_function()) {read_popen}

PR #1586, opened 2019-03-11
add function execute_shell_eval()

time-killer-games:
to make matters easier this replaces my previous pr.

two things I need to know from Josh:

- do you want me to cout the error info or would you prefer it to go as a second return value, which i don't know how to do? In debug mode, I could even make it show in a graphic error message.

- does this work for multiple lines in the command argument? or do i need to std::replace each new line and carriage return with a whitespace? I'm still not 100% sure whether even that would work tbh.

thanks babe!

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -10,7 +10,7 @@
 #include <errno.h>
 #include <stdint.h>
 
-ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
+static inline ssssize_t getline(char **lineptr, size_t *n, FILE *stream) {
   size_t pos;
   int c;
 
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -162,7 +162,7 @@ unsigned long get_timer() {  // microseconds since the start of the game
 std::string shellscript_evaluate(std::string command) {
   char *buffer = NULL;
   size_t buffer_size = 0;
-  std::string result = "";
+  std::string result;
 
   FILE *file = popen(command.c_str(), "r");
   while (getline(&buffer, &buffer_size, file) != -1)
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -7,56 +7,55 @@
 #include <cstdlib>         //free,malloc,realloc
 
 #ifdef _WIN32
-static inline ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
-  char *bufptr = NULL;
-  char *p = bufptr;
-  ssize_t size;
+#include <errno.h>
+#include <stdint.h>
+
+ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
+  size_t pos;
   int c;
 
-  if (lineptr == NULL && n == 0) {
+  if (lineptr == NULL || stream == NULL || n == NULL) {
+    errno = EINVAL;
     return -1;
   }
-  if (stream == NULL) {
-    return -1;
-  }
-  if (n == NULL) {
-    return -1;
-  }
-  bufptr = *lineptr;
-  size = *n;
 
   c = fgetc(stream);
   if (c == EOF) {
     return -1;
   }
-  if (bufptr == NULL) {
-    bufptr = malloc(128);
-    if (bufptr == NULL) {
+
+  if (*lineptr == NULL) {
+    *lineptr = malloc(128);
+    if (*lineptr == NULL) {
       return -1;
     }
-    size = 128;
+    *n = 128;
   }
-  p = bufptr;
+
+  pos = 0;
   while(c != EOF) {
-    if ((p - bufptr) > (size - 1)) {
-      size = size + 128;
-      bufptr = realloc((void *)bufptr, size);
-      if (bufptr == NULL) {
+    if (pos + 1 >= *n) {
+      size_t new_size = *n + (*n >> 2);
+      if (new_size < 128) {
+        new_size = 128;
+      }
+      char *new_ptr = realloc(*lineptr, new_size);
+      if (new_ptr == NULL) {
         return -1;
       }
+      *n = new_size;
+      *lineptr = new_ptr;
     }
-    *p++ = c;
+
+    ((unsigned char *)(*lineptr))[pos ++] = c;
     if (c == '\n') {
       break;
     }
     c = fgetc(stream);
   }
 
-  *p++ = '\0';
-  *lineptr = bufptr;
-  *n = size;
-
-  return p - bufptr - 1;
+  (*lineptr)[pos] = '\0';
+  return pos;
 }
 #endif
 
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -159,7 +159,7 @@ unsigned long get_timer() {  // microseconds since the start of the game
   return (enigma::time_current.tv_sec) * 1000000 + (enigma::time_current.tv_nsec / 1000);
 }
 
-std::string shellscript_evaluate(std::string command) {
+std::string execute_shell_eval(std::string command) {
   char *buffer = NULL;
   size_t buffer_size = 0;
   std::string result;
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -10,7 +10,7 @@
 #include <errno.h>
 #include <stdint.h>
 
-static inline ssssize_t getline(char **lineptr, size_t *n, FILE *stream) {
+static inline ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
   size_t pos;
   int c;
 
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -3,7 +3,8 @@
 #include <time.h> //CLOCK_MONOTONIC
 #include <sys/types.h>     //getpid
 #include <unistd.h>        //usleep
-#include <cstdio>          //popen, getline
+#include <cstdio>          //popen,getline
+#include <cstdlib>         //free
 
 namespace enigma {
 
--- a/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
@@ -40,7 +40,7 @@ static inline ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
   while(c != EOF) {
     if ((p - bufptr) > (size - 1)) {
       size = size + 128;
-      bufptr = realloc(bufptr, size);
+      bufptr = realloc((void *)bufptr, size);
       if (bufptr == NULL) {
         return -1;
       }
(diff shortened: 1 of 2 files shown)


ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
```diff
@@ -103,6 +105,24 @@ unsigned long get_timer() {  // microseconds since the start of the game
   return (enigma::time_current.tv_sec) * 1000000 + (enigma::time_current.tv_nsec / 1000);
 }
 
+std::string shellscript_evaluate(std::string command) {
+  char *buffer = NULL;
+  size_t buffer_size = 0;
+  std::string result = "";
```
JoshDreamland: This is the default construction of string; no need to specify it.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
```diff
@@ -103,6 +105,24 @@ unsigned long get_timer() {  // microseconds since the start of the game
...
+
+  FILE *file = popen(command.c_str(), "r");
+  while (getline(&buffer, &buffer_size, file) != -1)
+    result += buffer;
+
+  free(buffer);
+  pclose(file);
+
+  if (result.back() == '\n')
```
JoshDreamland: I don't think modifying the output is wise. If there was no newline in the output, `fgets`/`getline` will not add one.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
```diff
@@ -103,6 +105,24 @@ unsigned long get_timer() {  // microseconds since the start of the game
   return (enigma::time_current.tv_sec) * 1000000 + (enigma::time_current.tv_nsec / 1000);
 }
 
+std::string shellscript_evaluate(std::string command) {
+  char *buffer = NULL;
+  size_t buffer_size = 0;
+  std::string result = "";
+
+  FILE *file = popen(command.c_str(), "r");
+  while (getline(&buffer, &buffer_size, file) != -1)
```
JoshDreamland: This is not the C++ `getline` I thought you'd be using. This is a new routine added to the C standard. It will allocate a buffer for you of the correct size, but you must still free that buffer or you're just leaking memory.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
```diff
@@ -103,6 +105,24 @@ unsigned long get_timer() {  // microseconds since the start of the game
   return (enigma::time_current.tv_sec) * 1000000 + (enigma::time_current.tv_nsec / 1000);
 }
 
+std::string shellscript_evaluate(std::string command) {
```
JoshDreamland: You need to declare this in the header if people are going to use it. Also, consider an `execute_shell`-based name for discoverability.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
time-killer-games: I did free it. Did you not see it?

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
time-killer-games: well from my testing it *does* add a new line. I suppose this bullshit is an OS bug too? Also happens on Mac.... .. ... .

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
time-killer-games: Yeah i saw a question on stack overflow and the answer said std::getline was the C++ alternative to fgets. However i saw a different question that said std::getline can't be used with popen, but getline (the C function) could.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
time-killer-games: well technically evaluate is more of a correct term than execute as those are two different things but yeah i agree that's probably a good idea

time-killer-games: how's execute_shell_eval()?

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
time-killer-games: i know the linux man page claims it doesn't add one if it isn't found but i think it is not refering to the exclusion of the end of file where yes there is always a new line. As Robert said, that's how DOS and Unix generally works.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
time-killer-games: i know the Linux man page claims it doesn't add one if it isn't found but i think it is not referring to the exclusion of the end of file where yes there is always a new line. As Robert said, that's how DOS and Unix generally works.

time-killer-games: A lot of people do not typically like a random new line at the end of a string return value when it isn't asked for.

time-killer-games: I will be re-using this as a helper function with the new linux widget systems re-writes. So it is important that is does everything i need it to within reason

time-killer-games: @JoshDreamland could you tell me what purpose it serves to have a new line at the end, whether the function is responsible for it or otherwise?

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
JoshDreamland: Oh, no; I did miss that `free` line, actually.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
JoshDreamland: No OS bugs, here. The newline is annoying in some programs, and crucial in others. I believe we already have a `string_strip` utility method, and if not, we could really use one. Users who didn't want to deal with extra whitespace would just pass the result to `string_strip()`, while users who needed it (eg, as a delimiter) would use it however they needed.

Perhaps the circumstances where the newline is important are rare, particularly in game design (where the need for this function is, itself, rare), but you writing these two lines for the majority of users makes it impossible for the other users to determine whether there was or was not a newline.

Imagine someone ran `cat` or `unrar` and the output bytes encoded for a file. If that file ended in a newline, you would ruin the file for them in a way that no one would ever be able to figure out. Text files always end with newlines in Linux. If a binary file's last byte were a newline, you could render the file unusable until someone discovered what had happened and why.

ENIGMAsystem/SHELL/Platforms/General/POSIX/Utility.cpp
```diff
@@ -3,6 +3,62 @@
 #include <time.h> //CLOCK_MONOTONIC
 #include <sys/types.h>     //getpid
 #include <unistd.h>        //usleep
+#include <cstdio>          //popen,getline,fgetc,pclose
+#include <cstdlib>         //free,malloc,realloc
+
+#ifdef _WIN32
+static inline ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
```
JoshDreamland: Please don't paste code of a different license into arbitrary ENIGMA files. Actually, please don't declare this here at all—I'm unsurprised it's not supported across all systems; it's a C function as of 2010, and probably shouldn't be used in C++ at all. I didn't mind it given that Clang/GCC tend to support it, but evidently, that's not always true. Please just use `fread()` instead of attempting to read the output line-by-line.

time-killer-games: The license was public domain according to the comment. Which is perfectly acceptable last I checked.

time-killer-games: tread seems to be more limiting as there is no way to dynamically resize the input buffer, unlike getline.

time-killer-games: The proof is in the razzle frazzle doop dop jello pudding - https://stackoverflow.com/questions/735126/are-there-alternate-implementations-of-gnu-getline-interface/735472#735472

time-killer-games: I just updated It, this version is better: https://stackoverflow.com/questions/735126/are-there-alternate-implementations-of-gnu-getline-interface/47229318#47229318

time-killer-games: I feel like you guys a lot of the time will just tell me to make nonsense changes to push me around, being incredibly picky, down to the point of being sub-retarded.

time-killer-games: This is exactly why you guys make zero progress with this project for months at a time. None of you agree on how code should be written and instead of getting perfectly acceptable code merged we sit back and argue about 50,000 changes that don't need to happen.

JoshDreamland: The changes *do* need to happen, and at no point was your code "perfectly acceptable." At its *best*, your code didn't work with newlines, which you refused to acknowledge because you found a workaround for *your* use case with a non-newline character. As it appears now, your solution requires this 50-line helper function that doesn't build on Windows. And you're assuming it works because someone on StackOverflow pointed out that the last solution you copied from that site didn't work and they fixed it.

To be clear: this repository is not a shrine to copy-pasted segments of StackOverflow (even the revised, revised, revised copy that doesn't explode under the enclosed list of circumstances). Copying code from other people is **not** a sound approach to problem solving or tool authorship. It isn't my preference that you continue to iterate so frequently on pull requests like this. When you do, it ties up tons of CI resource, slows everyone else's development, and in general takes time from others to point out all the problems in your copy-pasted code. We tolerate it because your persistence is admirable, and in general, it's in everyone's best interests to help you and other contributors learn. But this is not a good approach to learning. You haven't taken a step back and realized that your line-based solution is *not* fruitful, in spite of my gently pointing it out multiple times. You won't succeed this way; you will keep making a bigger and bigger mess of more and more lines of foraged code.

I *could* write this method myself, but I don't think that would really help anyone. It's roughly six lines of code using `fread`, but you seem averse to it, or incapable of using it. I'm **not** going to accept any old solution simply because it fits some use case you have in mind. Not when it's broken, for obvious and easily-reproduced reasons, and ten times as many lines as necessary. Even if you get the AppVeyor build passing, I'm not going to merge in a 50-line method that no one (***including*** you) needs, that you copied off of StackOverflow.

If you would like me to write this method for you, I am willing to do so. In fact, any time you would like a new method, feel free to file a bug. They usually get filled quickly (a notable exception was `keyboard_key_press` on Linux, because it has dependencies on libxdo, which is bonkers to install).

I appreciate your effort, but again, this is *not* the way to attempt to implement a function.

time-killer-games: Well, if you could write it at this point, that would be great. I spent my whole day on this. You would've spent a few minutes. I need this for a lot of code I'm about to add to our repository, and hugar also needs it for what he is looking to do. The reason I "refused" to use the function you suggestion, I already had the impression getline() was ok with you, it doesn't really matter that my impression was a misunderstanding or if that really was what you meant even though it wasn't. Because of what I thought you meant, I spent all of today from *before* the crack of dawn to right now after 11pm, working on trying to figure out that getline function, and *then* you tell me to spend that kind of time all over again *still* refusing to tell me how to use functions you spit out the names of with no other details barely. and *then* assume code copied from stack overflow is broken without even looking at it, simply because you don't like me copying and pasting shit even though I've where Robert and fundies get their code for instance Robert's get_directory_alt() function was taken from codeguru or whatever it's called and the root argument was done completely wrong to the point of being useless.

time-killer-games: Sorry I know you probably did look at the code enough to know it's broken, but still, the other things I said remain. I'm not upset anymore, I just needed to erupt the mega volcano of rage and now I'm calm.

time-killer-games: yeah i am not going to spend another 12 hours on this. I'd appreciate a code snippet, and if you really want me to learn, you can briefly explain what it does and why it works.

JoshDreamland:
See #1590.

PR #1637, opened 2019-03-19
Add new functions to Cocoa Widgets

time-killer-games:
**New Functions Added:**

- show_message_cancelable(str) same as show_message() but has a cancel button.
- show_question_cancelable(str) same as show_question() but has a cancel button.
- show_attempt(str) shows a special type of error message, that unlike show_error(), it does not provide a way to abort the application. Instead, it has a 'Retry' and a 'Cancel' button.
- message_get_caption() returns a custom message caption that was previously set by the developer, if one was set. Otherwise, it will return an empty string.
- message_set_caption(str) sets a custom message caption for the message, question, attempt, error, and input, and hidden input dialogs. Set str to an empty string to restore the default value, which is either the game window title or "Error" depending on the dialog type.

**Button Return Values:**

- 'X' close button may vary in return value. This behavior is relative to the widget system in use and can not be changed with Zenity or KDialog due to limited features. macOS has no close buttons.
- 'Abort' aborts the app; no return value.
- 'OK' and 'Yes' return 1.
- 'No' and 'Retry' return 0.
- 'Cancel' and 'Ignore' return -1.

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
@@ -86,7 +86,7 @@ int show_message(const string &str) {
   return cocoa_show_message(str.c_str(), false, dialog_caption.c_str());
 }
 
-int show_message_cancelable(const string &str) {
+int show_message_cancelable(string str) {
   if (dialog_caption == "") dialog_caption = cocoa_dialog_caption();
   return cocoa_show_message(str.c_str(), true, dialog_caption.c_str());
 }
--- a/ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
@@ -38,15 +38,20 @@ bool widget_system_initialize() {
   
 } // namespave enigma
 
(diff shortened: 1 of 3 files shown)


ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
```diff
@@ -76,34 +82,51 @@ void show_info(string text, int bgcolor, int left, int top, int width, int heigh
...
+
+int show_attempt(string str) {
+  if (error_caption == "") error_caption = "Error";
+  return cocoa_show_attempt(str.c_str(), error_caption.c_str());
 }
 
 string get_string(string str, string def) {
-  string result = cocoa_input_box(str.c_str(), def.c_str());
-  return result;
+  if (dialog_caption == "") dialog_caption = cocoa_dialog_caption();
```
JoshDreamland: In the future, please prefer `.empty()` to `== ""`.

time-killer-games: Will do. I can make that change before the merge if you wish. I'll leave it up to you.

time-killer-games: Is doing what I currently have going to potentially break in the future or something? I'm curious why it's prefered.

PR #1639, opened 2019-03-21
Remove Dependency on TFD

time-killer-games:
**New Functions Added:**

- show_message_cancelable(str) same as show_message() but has a cancel button.
- show_question_cancelable(str) same as show_question() but has a cancel button.
- show_attempt(str) shows a special type of error message, that unlike show_error(), it does not provide a way to abort the application. Instead, it has a 'Retry' and a 'Cancel' button.
- message_get_caption() returns a custom message caption that was previously set by the developer, if one was set. Otherwise, it will return an empty string.
- message_set_caption(str) sets a custom message caption for the message, question, attempt, error, and input, and hidden input dialogs. Set str to an empty string to restore the default value, which is either the game window title or "Error" depending on the dialog type.

**Button Return Values:**

- 'X' close button may vary in return value. This behavior is relative to the widget system in use and can not be changed with Zenity or KDialog due to limited features. macOS has no close buttons.
- 'Abort' aborts the app; no return value.
- 'OK' and 'Yes' return 1.
- 'No' and 'Retry' return 0.
- 'Cancel' and 'Ignore' return -1.

**Sample Test Project:** [dialogs.zip](https://github.com/enigma-dev/enigma-dev/files/3006667/dialogs.zip)


time-killer-games:
@JoshDreamland If you don't review this puppy right now, my dick -> your butthole. Understood?

ENIGMAsystem/SHELL/Widget_Systems/KDialog/dialogs.cpp
```diff
@@ -15,10 +15,523 @@
...
+string get_open_filenames(string filter, string fname) {
+  string str_command; string pwd;
+  string str_title = "Open";
+  string str_fname = filename_name(fname);
+
+  pwd = ""; if (str_fname.c_str() && str_fname[0] != '/' && str_fname.length()) pwd = string("\"$PWD/\"") +
+    string("\"") + add_escaping(str_fname, false, "") + string("\""); else pwd = "\"$PWD/\"";
+
+  str_command = string("kdialog ") +
+  string("--attach=$(xprop -root 32x '\t$0' _NET_ACTIVE_WINDOW | cut -f 2) ") +
```
JoshDreamland: There's no need for `string("") + string("")`; just use `string(""  "")`. In fact, the `string()` construction is unnecessary, here, anyway, since `pwd` is a string.

ENIGMAsystem/SHELL/Widget_Systems/KDialog/dialogs.cpp
```diff
@@ -15,10 +15,523 @@
...
+  string result = shellscript_evaluate(str_command);
+  return file_exists(result) ? result : "";
+}
+
+string get_open_filenames(string filter, string fname) {
+  string str_command; string pwd;
+  string str_title = "Open";
+  string str_fname = filename_name(fname);
+
+  pwd = ""; if (str_fname.c_str() && str_fname[0] != '/' && str_fname.length()) pwd = string("\"$PWD/\"") +
```
JoshDreamland: Also, this `pwd = ""` assignment doesn't actually do anything.

ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
```diff
@@ -64,9 +64,9 @@ static inline string remove_trailing_zeros(double numb) {
 
 void show_error(string errortext, const bool fatal) {
   #ifdef DEBUG_MODE
-  errortext = enigma::debug_scope::GetErrors() + "\n\n" + errortext;
+  errortext = enigma::debug_scope::GetErrors() + "\r\n\r\n" + errortext;
```
JoshDreamland: OSX doesn't care about `\r`, if I understand correctly.

time-killer-games: Yeah, I knew that already. It must have slipped my mind when I made that change.

time-killer-games: I believe the same is also true for Linux, iirc.

ENIGMAsystem/SHELL/Widget_Systems/KDialog/dialogs.cpp
time-killer-games: That I knew as well, in this case because you have told me this before. I must have missed that when I was revising.

time-killer-games:
@JoshDreamland you ready to merge?

PR #1653, opened 2019-03-27
Update math tests to ensure consistency with GML

9Morello:


CommandLine/testing/SimpleTests/math_test.sog/create.edl
```diff
@@ -55,14 +55,18 @@ gtest_assert_true(equal(-.5, f));
 gtest_assert_true(equal(d,   f));
 gtest_assert_true(equal(.1 + .1 + .1, .3));  // Epsilon is important, here!
 
+// Tests of equal that assure compatibility with GMS behavior
+gtest_assert_eq(1.0 + real("1.1"), 2.1);
```
JoshDreamland: Use `.125` instead of `.1`, or use the floating-point-friendly method (with the epsilon). Otherwise, equality is a matter of opinion from processor to processor.

PR #1660, opened 2019-04-04
Remove DlgMod because DLL calling is broken

time-killer-games:
with dll calling broken and no progress being made with that we can remove this as a temporary measure to not confuse users. we already have a broken GTK+ widget set, we don't need a any other broken widgets sets. there is no point, its broken until we fix dll calling, and people prefer not having a dll and will go with windows widgets anyway

- I also removed new lines added to debug scope in win32 show_error and moved it to debug scope's GetErrors so the new lines can be used in all widgets systems

I'm going to be updating Windows Widgets soon as a replacement for both DlgMod and the current Windows Widgets.


time-killer-games:
@JoshDreamland This one is ready for you to review. I'll let you know when the CI passes so you can mudge it.

time-killer-games:
> No need to delete things just change the about ey to represents none, build platforms none and it shouldn't appear as an option in lgm

I'll do what you suggested, if you find it completely necessary.

But I do want to clarify that I'm about to write a new version of Windows Widgets that will have all the same functions this one has and will be a long-awaited replacement for the two of them. The only difference between this and the Windows Widget is this is completely broken, and the new Windows Widgets won't be, because it won't rely on a third party DLL.

Does that make sense? imo we don't need more than one system for Windows and I think this is what I should've done to begin with was not create DlgMod, but instead update Windows Widgets to have UTF-8 support and all the other stuff Mac and Linux have.

fundies:
Just hide the broken one from users until you have a replacement ready. For all we know you could die tomorrow and then the next person wanting to write widgets would have no reference because you deleted it without ever sending a replacement.

time-killer-games:
Ok

PR #1661, opened 2019-04-06
Remove DlgMod visibility due to being broken

time-killer-games:
with dll calling broken and no progress being made with that we can remove this as a temporary measure to not confuse users. we already have a broken GTK+ widget set, we don't need a any other broken widgets sets. there is no point, its broken until we fix dll calling, and people prefer not having a dll and will go with windows widgets anyway


time-killer-games:
@fundies i did test this locally and DlgMod is still showing up on Windows. I don't know what that's all about. I also tried typing out the word "None" and all lower case "none" and those didn't do it either.

fundies:
Try putting "None" in both fields i mentioned. If that fails lgms just bugged so try deleting the ey file.

time-killer-games:
> Try putting "None" in both fields i mentioned. If that fails lgms just bugged so try deleting the ey file.

Yeah, I did that, again, and this time I also tried it both with and without the quotes (previously i just did without) and it didn't change anything.

I'll just delete the about file

PR #1662, opened 2019-04-07
jdi fixes

fundies:


fundies:
some fixes for msvc and mingw on newjdi branch

PR #1663, opened 2019-04-08
Unify draw_sprite_* implementations a little and force consistent formatting in that file.

YellowAfterlife:
As requested,


YellowAfterlife:
Tester: [test.gmx.zip](https://github.com/enigma-dev/enigma-dev/files/3053779/test.gmx.zip)
image

PR #1669, opened 2019-04-09
Nonsense

fundies:
possibly nonsense jdi fixes that cut error count by a lot.


CompilerSource/JDI/src/Parser/readers/read_type.cpp
```diff
@@ -336,7 +336,7 @@ int jdip::context_parser::read_referencers(ref_stack &refs, const full_type& ft,
         if (!d) return 1;
         
         if (token.type == TT_MEMBEROF) {
-          if (!(d->flags & DEF_CLASS)) {
+          if (!(d->flags & DEF_CLASS) && !(d->flags & DEF_TYPENAME)) {
```
JoshDreamland: This is somewhat dangerous; line 343 treats this as a `definition_class*`. That cast isn't valid if this original condition isn't true.

CompilerSource/compiler/jdi_utility.cpp
```diff
@@ -55,6 +55,11 @@ static void iterate_overloads(
        iter != d->overloads.end(); iter++) {
     visit_overload(iter->second, min, max, varargs_t);
   }
+
+  for (definition_template* templateOverload : d->template_overloads) {
+    if (min > templateOverload->params.size()) min = templateOverload->params.size();
```
JoshDreamland: This is actually the template parameter size; it doesn't actually correspond with the number of parameters for the function. The function will  be declared within the template scope pointed to by `templateOverload`.

CompilerSource/JDI/src/Parser/readers/read_type.cpp
fundies: so what's the safe way to handle this then? stl has template<typename _Ret, typename _Tp> but _Tp is a class not a type so I get ```Member pointer to non-class `_Tp'```

CompilerSource/compiler/jdi_utility.cpp
```diff
@@ -57,8 +57,10 @@ static void iterate_overloads(
   }
 
   for (definition_template* templateOverload : d->template_overloads) {
-    if (min > templateOverload->params.size()) min = templateOverload->params.size();
-    if (max < templateOverload->params.size()) max = templateOverload->params.size();
+    definition* def = templateOverload->def;
+    if (def->flags & DEF_OVERLOAD) {
+     visit_overload(static_cast<definition_overload*>(def), min, max, varargs_t);
+    }
   }
```
fundies: @JoshDreamland does this look more correct?

JoshDreamland: Yes, although I can't remember if it's going to contain an overload or a complete function. I recall spending effort on that, so maybe I did get it to be an overload, but it wouldn't surprise me if it were still a complete function that just happens to have one overload in it.

But anyway, if that code solves the problem, great, because it looks correct.

CompilerSource/JDI/src/Parser/readers/read_type.cpp
JoshDreamland: I would have to dig into how that other method is using its operand to know for sure. I'm not sure what it needs from a class that it can't get from a regular template parameter. The simplest short-term solution would be to revert this line, and have template parameters declared as `typename` or `class` actually extend `definition_class`.

PR #1676, opened 2019-04-14
Rewrite Variant With New Templates

JoshDreamland:
No idea if this works. My system is sort of being batshit right now. If it *doesn't* work, I think anyone should be basically capable of finishing this up, at this point.


fundies:
```template<typename T> enigma::StringFunc<T, bool>
operator!=(const T& a, const var &b) { return b != a; }``` will fix issues with mingw but the errors on the ci regarding clang seem indicative of a much larger issue of how you're doing this. I think you may be relying on gcc specific behavior regarding overloads.

RobertBColton:
Alright: abeee7b4698612dcfc9715cacf06fd568b56b734

We narrowed down the issue with Key to Success and Project Mario to the `room` variable being uninitialized. Hopefully that fixes EnigmaBot too and we can merge this!

RobertBColton:
Haha, and nobody will ever say var is slow again!

Var Benchmark

JoshDreamland:
What a saga. Thanks for your help, Robert.

PR #1739, opened 2019-06-02
allow using relative paths for ini writing

time-killer-games:
For reference:
https://stackoverflow.com/questions/49620394/writeprivateprofilestring-works-in-main-but-not-in-function

"Also, if you read their documentation, you would see that you SHOULD NOT use a relative path for the INI file, or else the file will be relative to the Windows installation folder, not your app folder."

GetFullPathName fixes this issue because the whole purpose of that function is to automatically convert relative paths to absolute.


JoshDreamland:
We should use an EDL library call for that, going forward, but this solves it for now. Thanks!

PR #1740, opened 2019-06-02
Fix using string as type/function.

JoshDreamland:
When we moved exclusively to namespace `enigma_user`, we didn't move toString there, so the function is no longer allowed in EDL. Problem is, `string()` was a macro to it, so that was necessarily broken.

The best option for now is to remove the macro and use just the type in EDL. EDL will not force the std::string constructor to succeed on objects passed to the method, so GCC can silently fix it in the preprocessor for the time being.

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/var4.h
+++ b/ENIGMAsystem/SHELL/Universal_System/var4.h
@@ -262,9 +262,11 @@ enum {
 std::string toString(const void*);
 std::string toString(double);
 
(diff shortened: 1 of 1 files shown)


JoshDreamland:
Seems the user who reported this was having other issues, but this also fixes the issue with string declarations (although tackily). We might consider still merging it.

JoshDreamland:
Let's submit this for now on the grounds that it makes things better. I'll deal with the fallout later.

PR #1755, opened 2019-06-19
Move Win32 INI Funcs to IniFileSystem Extension

time-killer-games:
I'm doing this so that on all platforms users will consistently be expected to enable the extension in order to use the INI functions.

Additionally, UTF-8 support is added.

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSfilemanip.cpp
@@ -37,75 +37,8 @@
 
 using namespace std;
 
-static std::string iniFilename = "";
-
 namespace enigma_user
 {
-
-void ini_open(std::string fname)
-{
-	char rpath[MAX_PATH];
-	GetFullPathName(fname.c_str(), MAX_PATH, rpath, NULL);
-	iniFilename = rpath;
-}
-
-void ini_close()
-{
-	iniFilename = "";
-}
-
-std::string ini_read_string(std::string section, std::string key, std::string defaultValue)
-{
-	char buffer[1024];
-	GetPrivateProfileString(section.c_str(), key.c_str(), defaultValue.c_str(), buffer, 1024, iniFilename.c_str());
-
-	return buffer;
-}
-
-float ini_read_real(std::string section, std::string key, float defaultValue)
-{
-	char res[255];
-	char def[255];
-	sprintf(def, "%f", defaultValue);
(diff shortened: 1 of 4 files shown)


time-killer-games:
@RobertBColton @JoshDreamland Alright, this one is ready for one of you two review! Thanks!

time-killer-games:
@JoshDreamland C'mon bro. This is not that big of a change or much code to review.

time-killer-games:
@RobertBColton after this gets merged I'm going to add UTF-8 support to our other file functions.

JoshDreamland:
Sam, UTF-8 support is inherent to a system unless it has logic to replace unknown characters. The problem with Windows is that it sees UTF-8 control sequences and replaces them with the encoding for the corresponding Latin-1 characters, which totally fucks them up. I really doubt that's the case with these INI functions. Have you tested reading and writing non-ascii characters?

time-killer-games:
@JoshDreamland I accidentally deleted an important partition to my primary Windows machine, and ended up replacing it with Kubuntu all together.

That said, on a different machine, I am installing ENIGMA as we speak to confirm your theory. But I doubt Microsoft would've provided W variants of the INI functions for no apparent reason. If you happen to be right, and it works with the A variants, then that really would make no sense to me.

I'll let you know what my results are once I've tested both reading and writing.

time-killer-games:
@JoshDreamland 

We still need UTF-8 support because of the file argument.

While this code:
```
var test;
ini_open("Тест.ini");
ini_write_string("Тест", "Тест", "Тест");
test=ini_read_string("Тест", "Тест", "Тест");
show_message(test)
ini_close();
```

Wrote this to file:
```
[Тест]
Тест=Тест
```

The file being read and written to was named "Ð¢ÐµÑÑ‚.ini".

In your face. :P

JoshDreamland:
If you must pass the filename as a wide string, that is fine; your widening each key-value pair is wasteful. Try those keys/values in the original logic.

time-killer-games:
@JoshDreamland not possible. All of the functions require the file argument. In order to use the wide filename argument, all of the arguments need to be wide. Please actually look at the source this time to see what I mean. The chart pointers needed to be widened either way otherwise there will be type conflicts and it won't build.

time-killer-games:
@JoshDreamland if you don't want to merge this for reasons you gave on discord, I might as delete the branch. You don't have to review right away. But when you see this message I would like to know where you stand on whether this change is even with you considering as a temporary measure. Because if you don't, we both have more important things to move on to in that case, like my other pr and all the stuff Robert is working on.

PR #1764, opened 2019-06-28
call sound_add with sound_replace if !sound_exists

time-killer-games:
this issue was reported by hugar.

fixes sound_replace to not need a sound that exists in order to use. helps users who don't want to create "dummy" sound resources in advance.


time-killer-games:
@JoshDreamland smallest change in the world. Please review when you get the chance. Thanks! Please note on Windows before this change the game would crash. Linux had no problems with the previous code.

time-killer-games:
@JoshDreamland thanks for pointing that out, I just realized that. Makes sense. This can be closed in that case.

PR #1777, opened 2019-07-06
Add function directory_contents().

time-killer-games:
Shows the contents of a directory. Not recursive.

I might add a recursive version in the future, and if i do, I'll name the function directory_contents_ext() with an additional boolean argument to set whether to be recursive and include all sub-directories for the return value.

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/estring.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/estring.cpp
@@ -24,16 +24,6 @@
 #include "var4.h"
 #include "estring.h"
 
-#include "Platforms/General/PFmain.h"
-#include "Platforms/General/PFfilemanip.h"
-
-using enigma_user::file_exists;
-using enigma_user::directory_exists;
-using enigma_user::filename_absolute;
-using enigma_user::file_find_first;
-using enigma_user::file_find_next;
-using enigma_user::file_find_close;
-
 #ifdef DEBUG_MODE
 #include "libEGMstd.h"
 #include "Widget_Systems/widgets_mandatory.h"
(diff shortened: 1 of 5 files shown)


ENIGMAsystem/SHELL/Universal_System/estring.cpp
```diff
@@ -383,4 +393,27 @@ var string_split(const std::string &str, const std::string &delim,
   return res;
 }
 
+string directory_contents(string dname) {
+  if (string_replace_all(dname, " ", "") == "") dname = ".";
+  if (directory_exists(dname)) {
+    string rpath = filename_absolute(dname);
+    #if CURRENT_PLATFORM_ID == OS_WINDOWS
```
JoshDreamland: This is why we need `file_path_join`. Doing operating system checks in every file in the engine isn't good.

ENIGMAsystem/SHELL/Universal_System/estring.cpp
```diff
@@ -383,4 +393,27 @@ var string_split(const std::string &str, const std::string &delim,
...
+    if (rpath.back() != '\\') rpath += "\\";
+    #else
+    if (rpath.back() != '/') rpath += "/";
+    #endif
+    string item = file_find_first(rpath + "*", fa_readonly + fa_hidden + fa_sysfile + fa_volumeid + fa_directory + fa_archive);
```
JoshDreamland: Use `|` to combine flags. Using `+` will cause overlapping flags to misbehave.

ENIGMAsystem/SHELL/Universal_System/estring.cpp
```diff
@@ -383,4 +393,27 @@ var string_split(const std::string &str, const std::string &delim,
...
+    string res = rpath + item;
+    while ((file_exists(rpath + item) || directory_exists(rpath + item)) && item != "") {
+      item = file_find_next();
+      if (item != "." && item != ".." && item != "")
+        res += "\n" + rpath + item;
+    }
+    if (res == rpath + item) return "";
```
JoshDreamland: Returning here will leave the `file_find` open. Also, why do you want to return the empty string when a directory only contains one item?

ENIGMAsystem/SHELL/Universal_System/estring.cpp
```diff
@@ -383,4 +393,27 @@ var string_split(const std::string &str, const std::string &delim,
...
+  if (directory_exists(dname)) {
+    string rpath = filename_absolute(dname);
+    #if CURRENT_PLATFORM_ID == OS_WINDOWS
+    if (rpath.back() != '\\') rpath += "\\";
+    #else
+    if (rpath.back() != '/') rpath += "/";
+    #endif
+    string item = file_find_first(rpath + "*", fa_readonly + fa_hidden + fa_sysfile + fa_volumeid + fa_directory + fa_archive);
+    string res = rpath + item;
+    while ((file_exists(rpath + item) || directory_exists(rpath + item)) && item != "") {
```
JoshDreamland: Why do you expect `file_find` to return a path that doesn't exist? By contract, it returns the empty string when iteration is complete.

time-killer-games: yeah idk man i wasn't thinking i guess

time-killer-games:
> I'm not sure this method helps anyone. It would be better to tweak the file_find method to correct the behavior you're trying to work around (returning `.` and `..`). Apart from that, this method won't save any lines.
> 
> If we did want this method, it would not belong in estring. And it should probably return the files as an array (var) rather than a string.

when importing files from a directory into a sandbox directory for editing, this comes in handy, rather than copying the the folder, copy it's contents so it won't be in a pointless subfolder in the destination directory. That's kinda one of this biggest reasons for this. Anyway, I think i fixed everything.

PR #1779, opened 2019-07-06
filename_join, fa_all, fa_files_all

time-killer-games:


ENIGMAsystem/SHELL/Platforms/General/POSIX/POSIXdirectory.cpp
```diff
@@ -34,8 +51,13 @@ std::string filename_absolute(std::string fname) {
   if (file_exists(fname) || directory_exists(fname)) {
     char rpath[PATH_MAX];
     char *ptr = realpath(fname.c_str(), rpath);
-    if (ptr != NULL) { return rpath; }
-  } return "";
+    if (ptr != NULL) return add_slash(ptr);
```
JoshDreamland: This function used to support canonicalizing filenames. Now it converts filenames into directories.

time-killer-games: @JoshDreamland fix it

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
```diff
@@ -466,9 +466,17 @@ std::string filename_absolute(std::string fname) {
   if (file_exists(fname) || directory_exists(fname)) {
     wchar_t rpath[MAX_PATH];
     tstring tstr_fname = widen(fname);
-    GetFullPathNameW(tstr_fname.c_str(), MAX_PATH, rpath, NULL); 
-    return shorten(rpath);
-  } else { return ""; }
+    GetFullPathNameW(tstr_fname.c_str(), MAX_PATH, rpath, NULL);
+    if (directory_exists(ptr)) return add_slash(shorten(rpath));
```
JoshDreamland: You're now checking for existence twice.

time-killer-games: While it isn't needed here, it is needed for realpath otherwise it would crash

time-killer-games: Anyway, I fixed it along with something you didn't point out that was also wrong with the code you referenced.

time-killer-games:
@JoshDreamland tested everything locally. I'm ready for you to start the ci.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
```diff
@@ -463,12 +463,19 @@ void execute_program(std::string fname, std::string args, bool wait) { execute_p
...
+  wchar_t rpath[MAX_PATH];
+  tstring tstr_fname = widen(fname);
+  GetFullPathNameW(tstr_fname.c_str(), MAX_PATH, rpath, NULL);
+  if (directory_exists(shorten(rpath))) return add_slash(shorten(rpath));
+  if (file_exists(shorten(rpath))) return shorten(rpath);
+  return "";
+}
+
+std::string filename_join(std::string prefix, std::string suffix) {
+  if (directory_exists(prefix))
```
JoshDreamland: Let's avoid this method doing I/O. The typical use case will be to pass the result of this operation directly to another function, so returning the empty string won't help.

ENIGMAsystem/SHELL/Platforms/General/POSIX/POSIXdirectory.cpp
```diff
@@ -31,11 +49,20 @@ bool set_working_directory(string dname) {
...
+  char *ptr = realpath(fname.c_str(), rpath);
+  if (ptr != NULL) {
+    if (directory_exists(ptr)) return add_slash(ptr);
+    if (file_exists(ptr)) return ptr;
+  }
+  return "";
+}
+
+std::string filename_join(std::string prefix, std::string suffix) {
+  if (directory_exists(prefix))
```
JoshDreamland: Same thing here.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
time-killer-games: fixed

ENIGMAsystem/SHELL/Platforms/General/POSIX/POSIXdirectory.cpp
time-killer-games: fixed

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
```diff
@@ -463,12 +463,17 @@ void execute_program(std::string fname, std::string args, bool wait) { execute_p
...
-  if (file_exists(fname) || directory_exists(fname)) {
-    wchar_t rpath[MAX_PATH];
-    tstring tstr_fname = widen(fname);
-    GetFullPathNameW(tstr_fname.c_str(), MAX_PATH, rpath, NULL); 
-    return shorten(rpath);
-  } else { return ""; }
+  if (string_replace_all(fname, " ", "") == "") fname = ".";
+  wchar_t rpath[MAX_PATH];
+  tstring tstr_fname = widen(fname);
+  GetFullPathNameW(tstr_fname.c_str(), MAX_PATH, rpath, NULL);
```
JoshDreamland: If this method fails, it will return zero. Regardless, it returns the length of the result string. It would be good to use that length to construct a tstring, ie,
```cpp
tstring result(rpath, GetFullPathNameW(tstr_fname.c_str(), MAX_PATH, rpath, NULL));
return shorten(result);
```

Since the system method already has to check disk to evaluate this path, we can assume the file exists when it completes, so checking if `file_exists` doesn't help us.

Also, why do we care if the string ends in a slash? You can refer to a directory with or without the slash.

time-killer-games:
f = file_find_first(working_directory + "/*.*", fa_archive | fa_directory | fa_hidden | fa_readonly | fa_sysfile | fa_volumeid);
i = f;

while (i != "") {
  i = file_find_next();
  f += "#" + i;
}

file_find_close();
show_message(f);

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
time-killer-games: I want all functions that return existing paths to include the trailing slash in case the user doesn't specify it. for example, our working_directory includes it:

`example = working directory + "file.png"`

If we didn't add the trailing slash, this code wouldn't work, and would also be incompatible with how GMS 1.4 and 2.x does things.

ENIGMAsystem/SHELL/Platforms/General/POSIX/POSIXfilemanip.cpp
```diff
@@ -83,9 +83,7 @@ string file_find_next() {
 
   const int not_attrib = ~fff_attrib;
 
-  if (r == "." or r == ".."                                                      // Don't return ./ and
-      or ((r[0] == '.' or r[r.length() - 1] == '~') and not_attrib & fa_hidden)  // Filter hidden files
-  )
+  if ((r[0] == '.' or r[r.length() - 1] == '~') and not_attrib & fa_hidden)  // Filter hidden files
```
JoshDreamland: Why did you change this?

time-killer-games: See screenshot I posted fucktard. I proved you wrong about like 3 things today. Believe it or not, there are cases where that .. and . are very useful. One use would be to make your own custom gui alternative for get_open_filename skinned with your own sprites rather than opening an external window with a native look. This would allow adding a button that links to the parent folder if one exists.

JoshDreamland: I will say this here, as well. Your attitude on this matter is NOT acceptable. Just because Windows as Game Maker share a behavior does not mean that ENIGMA should have it. We seek to be useful as a game development platform. Not to simulate glitches and nuances in 5-year-old, 10-year-old, or 50-year-old APIs.

This isn't some game where you say "I proved you wrong!!" and now you win. Regardless of who has learned what about these APIs, today, it is NOT RELEVANT to this discussion how they behave. We should be looking at what it means to build a good toolkit.

If a project needs . and .. for some reason, it's easy to tack them on. They are unreliable in these systems, and are generally hard to remove. It should not be the default behavior. Your PR yesterday showed a great example of the usual case; a user's first order of business being to continue iteration as soon as . or .. are encountered.

I would not be strictly opposed to adding a flag to include them, but I don't value the behavior and would sooner get rid of it entirely.

PR #1785, opened 2019-07-09
[win32, zenity] fix dialog icons, [kdialog] fix error message in file dialogs

time-killer-games:
makes win32 and zenity have the correct icons in the dialog and, in zenity's case, running in desktops using KDE, the title bar as well. OK/Cancel should have a question mark instead of "i" for information.

error happened with the default kdialog shipped with certain linux distro's (since disco dingo release), and not the one you would've installed from the now obsolete package kde-baseapps-bin. This fixes the issue.

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/Zenity/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/Zenity/dialogs.cpp
@@ -108,23 +108,21 @@ static int show_message_helperfunc(string message) {
     dialog_caption = window_get_caption();
 
   string str_command;
-  string str_title;
-  string str_cancel;
+  string str_title = add_escaping(dialog_caption, true, " ");
+  string str_cancel = "--info --ok-label=OK ";
+  string str_icon = "\" --icon-name=dialog-information);";
   string str_echo = "echo 1";
 
-  if (message_cancel)
+  if (message_cancel) {
     str_echo = "if [ $? = 0 ] ;then echo 1;else echo -1;fi";
-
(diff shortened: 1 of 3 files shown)


time-killer-games:
@RobertBColton @JoshDreamland could one of you merge this when you get the chance? This has been tested locally and all checks have passed. It's a very subtle change. Thanks!

ENIGMAsystem/SHELL/Widget_Systems/Zenity/dialogs.cpp
```diff
@@ -108,23 +108,21 @@ static int show_message_helperfunc(string message) {
...
-  if (message_cancel)
     str_cancel = "--question --ok-label=OK --cancel-label=Cancel ";
+    str_icon = "\" --icon-name=dialog-question);";
+  }
 
   str_command = string("ans=$(zenity ") +
   string("--attach=$(sleep .01;xprop -root 32x '\t$0' _NET_ACTIVE_WINDOW | cut -f 2) ") +
   str_cancel + string("--title=\"") + str_title + string("\" --no-wrap --text=\"") +
-  add_escaping(message, false, "") + string("\" --icon-name=dialog-information);") + str_echo;
+  add_escaping(message, false, "") + str_icon + str_echo;
```
JoshDreamland: If these commands get much more complicated, please consider removing the quotes and brackets from `str_icon`, and any other variables you add like this. It's better to keep parentheses and semicolons in view of the command being assembled. Putting both quotes in the variable would make sense, but putting just one quote/parenthesis to match the same in a previously-appended string variable? That gets dangerous quickly.

PR #1811, opened 2019-07-18
Kill POSIX crap in Win32 execute_shell_for_output

time-killer-games:
I also intend to rewrite fileio.cpp for Win32 that doesn't use POSIX API calls and supports UTF-8, in a different pr.

`show_message(execute_shell_for_output('cmd /c dir c:\'));`

Capture


time-killer-games:
@JoshDreamland ready for review.

time-killer-games:
@JoshDreamland Note for review: the Posix version of execute_shell_for_output() will always return a new line at the end of the string no matter what. To adopt the same behavior, I did the windows equivalent by checking the function for failure scenarios, and under those conditions, return a carriage return and new line instead of an empty string, so that new line will always be at the end of the string regardless, just as the Posix function does (however there are no conditions the Posix function will fail, as even an invalid argument specified will output a lone new line character for the output string). I just wanted to explain why I did that; no reason other than consistency.

time-killer-games:
@JoshDreamland now passing! Review when you can and thank you! :D

JoshDreamland:
Nicely done.

PR #1822, opened 2019-07-21
[XLib] Get Primary Monitor Display Size like Windows/SDL

time-killer-games:
Just doing what the other platforms already do for consistency's sake. Will add optional enum argument to measure all monitors combined in a different pr. I need @JoshDreamland to test this on his monitors.
`draw_text(32, 32, string(display_get_width()) + "x" + string(display_get_height()));`
Screenshot_20190721_061318


time-killer-games:
@JoshDreamland hey boy

PR #1849, opened 2019-08-05
Emake Fonts

fundies:
This PR adds support for emake fonts.

DONE:

*  emake can now load fonts
* All fonts now go through the same code path regardless of if they're from the gmx, or added later using font_add / font_add_sprite
* LGM also now uses the same path as emake in-order to keep consistent rendering between builds
* I've fixed a regression where enigma was no longer compressing the font data
* All font data (excluding image data) is now written to source instead of tacked on the exe

TODO:
* I've added font finding functions but they depend on c++17 and could use more rigorous testing 
* I've put a framework in place to load gmx fonts with pre-rendered texture atlases but I have no idea how the gmx attributes translate to our GlyphMetrics struct 
* font_add() and font_add_sprite() are completely untested and likely broken
* I've added a generic interface to the rect_packer that should be applied in texture_atlas.cpp
*  We should be able to use one generic function for loading all compressed data on exe. I started one in fontinit.cpp but perhaps that can be continued in another PR
* We probably want to include some font to use as a default as only windows users are likely to have arial.  We can grab it via install.sh or whatever.
* It seems my lastest changes to AssetArray have broken the default font

NOTE: 
The compiler is now dependent on freetype and c++17 in-order to load fonts properly

### DIFF
--- a/CompilerSource/backend/GameData.cpp
+++ b/CompilerSource/backend/GameData.cpp
@@ -89,6 +89,35 @@ ImageData loadImageData(const std::string &filePath, int &errorc) {
   return ImageData(pngwidth, pngheight, data, dataSize);
 }
 
+static bool font_load_helper(const buffers::resources::Font& font, ImageData& data, RawFont& raw_font) {
+  int error = 0;
+
+  //FIXME: should be has_image() but lgm writter bugged and always writes a image even if one doesnt exist
+  if (font.glyphs_size() > 0) { // font with pre-rendered texture atlas
+    data = loadImageData(font.image(), error);
+    if (error) return false; // image file missing
+  } else { // load ttf and generate texture atlas
+    std::string ttf = enigma_user::font_find(font.font_name(), font.bold(), font.italic(), false);
+    if (!ttf.empty()) {
+      std::cout << "Loading font file: " << ttf << std::endl;
+      for (const buffers::resources::Font::Range& range : font.ranges())
+        raw_font.ranges.emplace_back(range.min(), range.max());
+      
+      if (!font_load(ttf, font.size(), raw_font)) return false; // ttf load error
+      
+      unsigned textureW, textureH;
+      unsigned char* pxdata = font_pack(raw_font, textureW, textureH);
+      int size = textureW * textureH;
+      unsigned char* cpxdata = zlib_compress(pxdata, size);
+      delete[] pxdata;
+      data = ImageData(textureW, textureH, cpxdata, size);
+
+    } else return false; // font load error
+  }
+
+  return true;
+}
+
 struct ESLookup {
   struct Lookup {
     std::string kind;
@@ -194,30 +223,15 @@ FontData::FontData(const deprecated::JavaStruct::Font &font):
   data.set_bold(font.bold);
   data.set_italic(font.italic);
 
-  for (size_t i = 0; i < font.glyphRangeCount; ++i) {
-    GlyphRange range(font.glyphRanges[i].rangeMin, font.glyphRanges[i].rangeMax);
-    raw_font.ranges.emplace_back(range);
-    for (size_t j = 0; j < range.size(); ++j) {
-      deprecated::JavaStruct::Glyph& jg = font.glyphRanges[i].glyphs[j];
-      RawGlyph rg;
-      rg.character = font.glyphRanges[i].rangeMin + j;
-      rg.pxdata = jg.data;
-      rg.index = j;
-      rg.horiBearingX = 0;
-      rg.horiBearingY = 0;
-      rg.dimensions = enigma::rect_packer::pvrect(0, 0, jg.width, jg.height, -1);
-      rg.linearHoriAdvance = jg.advance;
-      rg.range = &raw_font.ranges.back();
-      raw_font.glyphs.emplace_back(rg);
-    }
+  for (int i = 0; i < font.glyphRangeCount; ++i) {
+    buffers::resources::Font_Range* range = data.add_ranges();
+    range->set_min(font.glyphRanges[i].rangeMin); 
+    range->set_max(font.glyphRanges[i].rangeMax);
   }
 
-  unsigned textureW, textureH;
-  unsigned char* pxdata = font_pack(raw_font, textureW, textureH);
-  int size = textureW * textureH;
-  unsigned char* cpxdata = zlib_compress(pxdata, size);
-  delete[] pxdata;
-  image_data = ImageData(textureW, textureH, cpxdata, size);
+  if (!font_load_helper(data, image_data, raw_font))
+    std::cerr << "Failed to load font: " << font.name << " aka "  << font.fontName << std::endl;
+  
 }
 
 PathData::PathData(const buffers::resources::Path &q, const std::string& name):
@@ -230,7 +244,7 @@ PathData::PathData(const deprecated::JavaStruct::Path &path):
   data.set_closed(path.closed);
   data.set_precision(path.precision);
 
-	// Discarded: backgroundRoomId;
+  // Discarded: backgroundRoomId;
   data.set_hsnap(path.snapX);
   data.set_vsnap(path.snapY);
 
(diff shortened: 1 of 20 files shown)


shared/fonts/fonts.cpp
```diff
@@ -0,0 +1,295 @@
...
+    } else {
+      for (auto& p: fs::directory_iterator(fs::path(path)))
+        font_add_search_path_helper(p.path());
+    }
+  } else DEBUG_MESSAGE("Freetype warning searching non-existing directory: " + path, MESSAGE_TYPE::M_WARNING);
+  
+  #endif
+}
+
+std::string font_find(std::string name, bool bold, bool italic, bool exact_match) {
```
fundies: this needs more testing

CompilerSource/compiler/components/write_font_info.cpp
```diff
@@ -45,30 +46,77 @@ int lang_CPP::compile_writeFontInfo(const GameData &game)
...
         << font->bold()      << ", "      // bool bold;
-        << font->italic()    << ", "      // bool italic;
-        << font.normalized_ranges.size()  // int glyphRangeCount;
-        << "}," << endl;
+        << font->italic()    << ", ";     // bool italic;
+
+        std::stringstream ranges;
+        ranges << "{ "; //begin glyph ranges
+        //FIXME: should be has_image() but lgm writter bugged and always writes a image even if one doesnt exist
+        if (font->glyphs_size() > 0)  { // font used pre-rendered texture
```
fundies: Add gmx image font support here

ENIGMAsystem/SHELL/Universal_System/Resources/fonts.cpp
```diff
@@ -0,0 +1,265 @@
...
+    if (character >= fgr.start && character < fgr.start + fgr.glyphs.size()) {
+      return fgr.glyphs[character - fgr.start];
+    }
+  }
+  return GlyphMetrics();
+}
+
+} // namespace fonts
+
+// TODO: move this
```
fundies: I wanted to create some LoadLater type class but decided to halfass it for now

shared/fonts/fonts.cpp
```diff
@@ -0,0 +1,295 @@
...
+     std::string fontFamily(reinterpret_cast<char*>(family.string), family.string_len);
+     fontFamilies[fontFamily] = p;
+
+     std::cout << fontFamily << std::endl;
+  }
+}
+#endif
+
+void font_add_search_path(std::string path, bool recursive) {
+  fontSearchPaths.emplace_back(path); // TODO: disallow duplicate paths?
```
fundies: Might want to protect users from adding same path infinite times in step or something.

ENIGMAsystem/SHELL/Universal_System/Resources/fonts.cpp
```diff
@@ -0,0 +1,265 @@
...
+        rect.w = x;
+        rect.h = y;
+      }
+    }
+  }
+
+  return rect;
+}
+
+static SpriteFont font_sprite_helper(enigma::sprite* sprite, uint32_t first, bool prop, int sep) {
```
fundies: I haven't tested this at all yet

shared/fonts/fonts.cpp
```diff
@@ -0,0 +1,295 @@
...
+             rg.pxdata[index] = charBitmap.buffer[index];
+            }
+          }
+       }
+       
+       rawFont.glyphs[currOffset++] = rg;
+    }
+  }
+
+  // FIXME: idk how GM calculates the line height
```
fundies: IDK what to do here. I think GM does something like using max glyph height but freetype recommends a different value for line height.

RobertBColton: The GML function `string_height` implementation is probably the answer. We have that function in the engine so maybe look at it, I'm sure the frontend would be written to be consistent with the engine right? Yeah.

CompilerSource/backend/GameData.cpp
```diff
@@ -592,5 +614,19 @@ GameData::GameData(deprecated::JavaStruct::EnigmaStruct *es,
 
 GameData::GameData(const buffers::Project &proj, const EventData* /*events*/):
     filename("") {
-  FlattenProto(proj, this);
+  
+  buffers::Project copy = proj;
```
RobertBColton, line 618: Yikes this looks bad. I don't really want us to copy the _entire_ game. Even if we do copy the entire game, are we sure this is the point in the build we want to do the copying? I'm actually not even sure if we already are copying the game before passing it to the RPC too. Josh had me make those parameters const for a reason.

Consider just removing the font after we are done. At the very least find how to make protobuf do a shallow copy only of the root tree node rather than the entire project.

RobertBColton:
I'm getting an error when building my Project Mario game from LGM on this branch after a clean rebuild of everything.
```
Writing executable information and resources.
Writing modes and settings
Writing object switch
Writing resource names and maxima
Thread-4:
java.lang.Error: Invalid memory access
        at com.sun.jna.Native.invokeInt(Native Method)
        at com.sun.jna.Function.invoke(Function.java:383)
        at com.sun.jna.Function.invoke(Function.java:315)
        at com.sun.jna.Library$Handler.invoke(Library.java:212)
        at com.sun.proxy.$Proxy0.compileEGMf(Unknown Source)
        at org.enigma.EnigmaRunner$CompilerThread.run(EnigmaRunner.java:674)

```

ENIGMAsystem/SHELL/Universal_System/colors.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+int merge_color(int c1,int c2,double amount)
+{
+  amount = amount > 1 ? 1 : (amount < 0 ? 0 : amount);
+  return (unsigned char)(fabs(COL_GET_R(c1)+(COL_GET_R(c2)-COL_GET_R(c1))*amount))
+  |      (unsigned char)(fabs(COL_GET_G(c1)+(COL_GET_G(c2)-COL_GET_G(c1))*amount))<<8
+  |      (unsigned char)(fabs(COL_GET_B(c1)+(COL_GET_B(c2)-COL_GET_B(c1))*amount))<<16;
+}
+
+}
```
RobertBColton, line 77: I wouldn't necessarily move all of this to Universal. Other graphics systems might want to use their own color formats that aren't BGRA like GM.

shared/texture.h
```diff
@@ -0,0 +1,18 @@
...
+  void init(); // loads a texture and deletes pxdata (engine only)
+  unsigned width;
+  unsigned height;
+  unsigned channels;
+  int ID;
+  unsigned char* pxdata;
+};
+} //namespace enigma
+
+#endif
```
RobertBColton, line 18: Uhm, I don't know how I feel about this. This is confusing because graphics texture vs raw image class you added. Can we not just use raw image here? Too bad we can't use our protos in the engine yet.

shared/fonts/fonts.cpp
fundies: pretty sure that will just return the line height it does here

CompilerSource/backend/GameData.cpp
fundies: The only other place I can think of to add this is in every other reader and I didn't want to put it in 5 places. If you can find a better place feel free to move it.

CompilerSource/backend/GameData.cpp
```diff
@@ -490,10 +503,19 @@ int FlattenTree(const buffers::TreeNode &root, GameData *gameData) {
       gameData->backgrounds.emplace_back(root.background(), root.name(), data);
       break;
     }
+    case TypeCase::kFont: {
+      ImageData data;
+      RawFont raw_font;
+      if (!font_load_helper(root.font(), data, raw_font)) {
+        std::cerr << "Failed to load font: " << root.name() << " aka " << root.font().font_name() << std::endl;
+        return -4; // font load error
```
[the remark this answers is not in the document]
fundies: josh returns -whatever for each failed resource. read above code.

ENIGMAsystem/SHELL/Universal_System/colors.cpp
fundies: then they can amend these to take other color formats... We don't need copies of them

shared/texture.h
fundies: This is used to load a texture later. I wrote it long before RawImage you could modify RawImage to have a load later I guess

shared/fonts/font_pack.cpp
```diff
@@ -0,0 +1,56 @@
...
+  if (!pack_rectangles(font.glyphs, textureW, textureH))
+    return nullptr;
+
+  // intialize blank image (the caller is responisible for deleting pxdata)
+  unsigned char* pxdata = new unsigned char[textureW * textureH * font.channels];
+  for (unsigned y = 0; y < textureH; ++y) {
+    for (unsigned x = 0; x < textureW; ++x) {
+      unsigned index = y * textureW + x;
+      for (unsigned i = 0; i < font.channels; ++i)
+        pxdata[index+1] = 0;
```
RobertBColton, line 19: As discussed on Discord, this is a typo, you meant to add the control variable i not the number one.

shared/fonts/font_pack.cpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
```
RobertBColton, line 21:
```suggestion
  memset(pxdata, 0, textureW * textureH * font.channels);
```

shared/fonts/font_pack.cpp
```diff
@@ -0,0 +1,56 @@
+#include "fonts.h"
```
RobertBColton, line 1:
```suggestion
#include <cstring> // memset

```

CompilerSource/backend/GameData.cpp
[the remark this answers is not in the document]
RobertBColton: So you mean he wants a number for each type that failed, why didn't he just return the type case then?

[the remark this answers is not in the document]
fundies: ask josh? I'm just following his code here

RobertBColton:
I found the crash is because of static libgcc in `CompilerSource/Makefile` apparently. That is causing freetype to corrupt protobuf's memory somehow.

PR #1853, opened 2019-08-20
Small Tweak

time-killer-games:
yes, *another* widget tweak...

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
@@ -18,16 +18,27 @@
 #include "dialogs.h"
 #include "Widget_Systems/widgets_mandatory.h"
 #include "Widget_Systems/General/WSdialogs.h"
+#include <X11/Xlib.h>
 #include <string>
 using std::string;
 
 namespace enigma {
 
-bool widget_system_initialize() {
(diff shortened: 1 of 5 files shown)


time-killer-games:

ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
```diff
@@ -18,17 +18,28 @@
...
+  Display *d = XOpenDisplay(NULL);
+  Atom aKWinRunning = XInternAtom(d, "KWIN_RUNNING", True);
+  bool bKWinRunning = (aKWinRunning != None);
+  XCloseDisplay(d);
+  return bKWinRunning;
+}
+
 bool widget_system_initialize() {
+  // Defaults to the GUI toolkit (GTK+/Qt) that matches Desktop Environment.
+  current_widget_engine = kwin_running() ? kdialog_widgets : zenity_widgets;
```
JoshDreamland: KDE running doesn't really guarantee that kdialog is available. If both Zenity and KDialog are available, then you should of course prefer whichever system is running, but odds are, the user has neither one installed. And if they do, they almost certainly only have one or the other. That said, this is a better guess than what we had before.

time-killer-games:
@JoshDreamland u mudge? This will make hugar very happy

PR #1856, opened 2019-09-15
fix segfaults with with execute_* functions

time-killer-games:
previously, execute_program would crash if you dragged the window by the title bar or clicked the close button over and over while a program is running with the wait argument true, and the game would also crash with execute_shell_for_output whenever trying to close a NULL handle, due to trying to run a nonexistent file or invalid command. This stuff is now fixed and no longer an issue,

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
@@ -454,8 +454,8 @@ std::string execute_shell_for_output(const std::string &command) {
         DispatchMessage(&msg);
       }
     }
-    if (hStdOutPipeWrite != NULL) { CloseHandle(hStdOutPipeWrite); }
-    if (hStdInPipeRead != NULL) { CloseHandle(hStdInPipeRead); }
+    CloseHandle(hStdOutPipeWrite);
+    CloseHandle(hStdInPipeRead);
     char buf[4096] = { };
     DWORD dwRead = 0;
     DWORD dwAvail = 0;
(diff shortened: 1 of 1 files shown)


ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
```diff
@@ -498,9 +497,16 @@ void execute_program(std::string operation, std::string fname, std::string args,
 
   //wait until a file is finished printing
   if (wait && lpExecInfo.hProcess != NULL) {
-    ::WaitForSingleObject(lpExecInfo.hProcess, INFINITE);
-    ::CloseHandle(lpExecInfo.hProcess);
+    while (WaitForSingleObject(lpExecInfo.hProcess, 0) == WAIT_TIMEOUT) {
```
JoshDreamland: This 0 causes the loop to busy-wait. Better to use 1, or even a larger value like 5.

time-killer-games: I disagree; doing that will make the window cursor turn into a loading cursor, and to me that is very undesirable behavior, and that is the only difference it made with my testing. The way I have is will allow you to click on the close button and will act like the button is clicked, but the game will not close because the process is waiting. This is what it does on Linux and is the correct behavior. I encourage you to actually test this before you conclude it's not correct.

time-killer-games: @JoshDreamland does this apply to this function only, or does it work differently when using a thread handle, like i did in execute_shell_for_ouput()?

JoshDreamland: Basically, anywhere a loop runs without sleeping, it will tie up a CPU thread. So if this code appears elsewhere and I missed it, it'll have the same problem. You can fix it by specifying a higher timeout (but not so high that the operating system will assume you've died), or by adding a call to `sleep`.

time-killer-games:
@JoshDreamland I remembered wrong about the new line thing I said in my previous pr I think. If the command isn't valid or doesn't print anything there is no new line printed.

time-killer-games:
@JoshDreamland now passing

PR #1868, opened 2019-11-21
Egmrebase

fundies:
fix includes


CommandLine/libEGM/gmx.cpp
```diff
@@ -338,6 +331,25 @@ void PackRes(std::string &dir, int id, pugi::xml_node &node, google::protobuf::M
       if (gmxName == "GMX_DEPRECATED")
         continue;
 
+      if (gmxName == "eventtype") {
```
JoshDreamland: I haven't figured out why this breaks things, yet, but this is clearly the product of a merge error. This logic already appears below. We have a duplicate handler for "action," as well. Mind deleting those and retrying? It shouldn't fix  this problem, but it should help us narrow it down.

To me, it's important to figure out why this change breaks it, because there may be a root cause I'm not seeing and will therefore trip again.

Speaking of, can we get an integration test that actually fails on this PR?

CompilerSource/backend/GameData.h
```diff
@@ -137,6 +137,9 @@ struct GameData {
   /// Name of project file. Used in metadata, and later, resource loading.
   std::string filename;
 
+  /// Event names
```
JoshDreamland: This is misleading... this is all known data about all events, not just event names.

CompilerSource/compiler/compile.cpp
```diff
@@ -249,7 +249,7 @@ int lang_CPP::compile(const GameData &game, const char* exe_filename, int mode)
   }
 
   //Read the types of events
-  event_parse_resourcefile();
+  event_parse_resourcefile(game.events);
```
JoshDreamland: And if events is empty? Do you expect people to just paste the default events.res into every game?

CommandLine/libEGM/gmk.cpp
```diff
@@ -916,8 +950,7 @@ std::unique_ptr<Object> LoadObject(Decoder &dec, int /*ver*/) {
       if (second == -1) break;
 
       auto event = object->add_events();
-      event->set_type(i);
-      event->set_number(second);
+      dec.postponeEventName(event->mutable_name(), i, second);
```
JoshDreamland: It's not clear to me how long this ref lives. Adding more events may invalidate it.

CommandLine/libEGM/gmx.cpp
fundies: Everything should fail on this pr. At least any test harness calls

CompilerSource/backend/GameData.cpp
```diff
@@ -512,7 +512,7 @@ GameData::GameData(deprecated::JavaStruct::EnigmaStruct *es): filename(es->filen
   cout << "Transfer complete." << endl << endl;
 }
 
-GameData::GameData(const buffers::Project &proj): filename("") {}
+GameData::GameData(const buffers::Project &proj): filename(""), events(proj.game().events()) {}
```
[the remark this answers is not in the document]
fundies: idk ask robert

CompilerSource/compiler/compile.cpp
fundies: afaik robert was reading events.res on file load

PR #1869, opened 2019-11-30
fix show_error()

time-killer-games:
fixes cases where show_error() doesn't abort/exit when it should.


time-killer-games:

ENIGMAsystem/SHELL/Widget_Systems/Cocoa/dialogs.cpp
```diff
@@ -83,6 +83,13 @@ void show_debug_message(string errortext, MESSAGE_TYPE type) {
     #ifndef DEBUG_MODE
     fputs(errortext.c_str(), stderr);
     #endif
+    if (type == MESSAGE_TYPE::M_FATAL_ERROR || 
+      type == MESSAGE_TYPE::M_FATAL_USER_ERROR)
+    #ifdef DEBUG_MODE
+      abort();
+    #else
+      exit(0);
```
JoshDreamland: Why are you trying to exit with status 0 outside of debug mode? That means the program finished successfully. Aborting due to a fatal error isn't success by any metric.

time-killer-games: What status should I use?

PR #1876, opened 2020-01-22
remove some of the duplicate logic in our makefiles

fundies:


CommandLine/emake/Makefile
```diff
@@ -16,18 +15,23 @@ else
...
 
-rwildcard=$(wildcard $1/$2) $(foreach d,$(wildcard $1/*),$(call rwildcard,$d,$2))
-SOURCES := $(call rwildcard,$(SRC_DIR),*.cpp)
+ifeq ($(TESTS), TRUE)
+	TARGET=../../emake-tests
+	SOURCES := $(call rwildcard, ../libEGM-test,*.cpp)
+	LDFLAGS += -lgtest -lpthread -lgtest_main
+else
+	TARGET = ../../emake
+  SOURCES := $(call rwildcard,$(SRC_DIR),*.cpp)
```
JoshDreamland: This line starts with two spaces instead of a tab. It's curious that make is tolerating this.

Default.mk
```diff
@@ -0,0 +1,57 @@
...
+DEPENDS = $(OBJECTS:.o=.d)
+
+ifdef SHARED_SRC_DIR
+CXXFLAGS += -I$(SHARED_SRC_DIR) $(addprefix -I$(SHARED_SRC_DIR)/, $(SHARED_INCLUDES))
+SOURCES += $(addprefix $(SHARED_SRC_DIR),$(SHARED_SOURCES))
+OBJECTS += $(addprefix $(OBJ_DIR)/shared/,$(SHARED_SOURCES:.cpp=.o))
+DEPENDS += $(addprefix $(OBJ_DIR)/shared/,$(SHARED_SOURCES:.cpp=.d))
+endif
+
+$(TARGET): $(OBJECTS)
```
JoshDreamland: I might suggest moving this into each makefile instead of putting it here, although I see it gives you foo-clean and DEFAULT_GOAL for free.

CommandLine/emake/Makefile
```diff
@@ -16,18 +15,23 @@ else
...
 
-PROTO_DIR := $(SHARED_SRC_DIR)/protos
-CXXFLAGS  := -I$(SRC_DIR) -I../../CompilerSource -I$(SHARED_SRC_DIR) -I$(PROTO_DIR)/.eobjs -std=c++11 -Wall -Wextra -Wpedantic -g
-LDFLAGS   := $(OS_LIBS) -L../../ -lProtocols -lprotobuf
+CXXFLAGS  += -I../../CompilerSource -I$(PROTO_DIR)
+LDFLAGS   += $(OS_LIBS) -L../../ -lProtocols -lprotobuf
 
-rwildcard=$(wildcard $1/$2) $(foreach d,$(wildcard $1/*),$(call rwildcard,$d,$2))
-SOURCES := $(call rwildcard,$(SRC_DIR),*.cpp)
+ifeq ($(TESTS), TRUE)
```
JoshDreamland: Losing TARGET would also let us just build these two as targets instead of using `TEST=true $(MAKE) ...`

fundies: Well we still build them as targets from the main makefile and thats the only one anyone should be calling really

CommandLine/emake/Makefile
fundies: ill fix this just in case

PR #1885, opened 2020-02-13
some hacks to make cmake files mingw freindly

fundies:


fundies:
@JoshDreamland merge this once ci done thx

PR #1895, opened 2020-02-27
[Win32] implement proper display_get_x/y() funcs

time-killer-games:
Just like our X11/Xlib implementation, this gets the x/y values for the primary monitor, except this time for Windows. More info here: https://docs.microsoft.com/en-us/windows/win32/gdi/multiple-monitor-system-metrics

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSwindow.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSwindow.cpp
@@ -318,11 +318,13 @@ void display_mouse_set(int x,int y) {
 }
 
 int display_get_x() {
-  return 0; // TODO
+  RECT rc; GetWindowRect(GetDesktopWindow(), &rc);
+  return rc.left;
 }
 
 int display_get_y() {
-  return 0; // TODO
+  RECT rc; GetWindowRect(GetDesktopWindow(), &rc);
+  return rc.top;
 }
 
 int display_get_width() {
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSwindow.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSwindow.cpp
@@ -318,13 +318,13 @@ void display_mouse_set(int x,int y) {
 }
 
 int display_get_x() {
-  RECT rc; GetWindowRect(GetDesktopWindow(), &rc);
-  return rc.left;
+  // Windows is different than our Unix-like platforms in that this value is always zero...
+  return 0;
 }
 
 int display_get_y() {
-  RECT rc; GetWindowRect(GetDesktopWindow(), &rc);
-  return rc.top;
+  // Windows is different than our Unix-like platforms in that this value is always zero...
+  return 0;
 }
 
 int display_get_width() {


time-killer-games:
@JoshDreamland @RobertBColton It's now passing and ready for review/merge. Thanks!

RobertBColton:
Don't you mean `display_get_dpi_x`?
https://docs.yoyogames.com/source/dadiospice/002_reference/windows%20and%20views/display_get_dpi_x.html

time-killer-games:
> Don't you mean `display_get_dpi_x`?
> https://docs.yoyogames.com/source/dadiospice/002_reference/windows%20and%20views/display_get_dpi_x.html

That's something completely different Robert. That would return the Dots Per Inch, which on Windows machines at 100% scaling is 96. What this pr implements is getting the x/y position of the primary monitor. If the primary monitor is not set to the top-left most monitor in your setup, It will be the width and/or height for the x/y position of every monitor's width/height combined that comes before it on the x/y axis.

time-killer-games:
@RobertBColton Keep in mind these are EDL functions that is not available in GameMaker and also are used as helper function within the enigma_user and enigma namespaces alike for various purposes, and window_center() being the main purpose. We need window_center() placing the window at the correct coordinates for uses with multiple monitor setups.

RobertBColton:
Oh

time-killer-games:
@RobertBColton I'm not in a position to test this right now, but from what I read in the docs it appears i might have a false impression of how this works, and the two functions might always return zero regardless. It's unclear to me because the docs say:

> The rectangle of the desktop window returned by GetWindowRect or GetClientRect is always equal to the rectangle of the primary monitor, for compatibility with existing applications. Thus, the result of
> C++
> 
> GetWindowRect(GetDesktopWindow(), &rc);
> 
> will be:
> C++
> 
> rc.left = 0; 
> rc.top = 0; 
> rc.right = GetSystemMetrics (SM_CXSCREEN); 
> rc.bottom = GetSystemMetrics (SM_CYSCREEN);

I'm not certain of rc.left/to = 0 is just an example code of what it might equal in some setups, or if that is the value it will always return. It seems to contradict itself in that the code example it gives seems to imply the latter, but in the description before that it says it will return relative to the rectangle of the primary monitor, which will not always equal zero. Or perhaps it does. Maybe the origin of the primary monitor is always zero because the primary monitor determines the origin of the other monitors relative to that... I find this all very confusing.

time-killer-games:
@RobertBColton @JoshDreamland this pr is useless YellowAfterlife verified a test for me and the origin of the primary monitor is always 0,0 on Windows. I'll change the code back to what it was but remove the `// TODO` comment since that's irrelevant now. It's interesting how Windows is so very different from our other platforms in this regard. On Linux, the primary monitor origin is sometimes greater than zero and the origin of any given monitor is never negative on Unix-like platforms, but on Windows it can sometimes be negative. Weird.

RobertBColton:
Might consider the Nvidia control panel.
Nvidia Control Panel Multi Monitor Setup

time-killer-games:
Ok, updated the code comment for clarification. @RobertBColton do you know for a fact using the NVIDIA control panel in preference over the Windows default is a game changer?

RobertBColton:
I don't think I really follow, I would expect the primary monitor to always have (0,0) for its position. Unless maybe underscan or overscan was being applied in the Nvidia control panel. Regardless, GetMonitorInfo would still be useful to get the display coordinates in a multi-monitor setup, though I understand GM hasn't really had any multi-monitor support. Perhaps may I suggest an overload like `display_get_x(int dispId)` be added to ENIGMA?

RobertBColton:
Perhaps also maybe you have not seen EnumDisplayMonitors which also returns this information.
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumdisplaymonitors

Maybe it's closer to the Unix/Linux functionality you are talking about.

PR #1899, opened 2020-03-07
Fix SDL Without Breaking Eveything Else

time-killer-games:
Screenshot_20200307_124350

PR #1903, opened 2020-03-08
Fixes drawing on android

fundies:
GLES doesn't support texture2d and doesn't do bgra->rgba conversion so this pr provides fixes for both. Also, SDL behaves strangely when the window isn't fullscreen on android so I just force fullscreen for now.

Please merge https://github.com/enigma-dev/enigma-android/pull/1 & #1899 before this

### DIFF
--- a/ENIGMAsystem/SHELL/Graphics_Systems/OpenGL-Common/textures.cpp
+++ b/ENIGMAsystem/SHELL/Graphics_Systems/OpenGL-Common/textures.cpp
@@ -80,7 +80,7 @@ int graphics_create_texture_custom(unsigned width, unsigned height, unsigned ful
 int graphics_create_texture(unsigned width, unsigned height, unsigned fullwidth, unsigned fullheight, void* pxdata, bool mipmap) {
   unsigned char* converted_pxdata = bgra2rgba((unsigned char*)pxdata, width, height);
   int ret = graphics_create_texture_custom(width, height, fullwidth, fullheight, converted_pxdata, mipmap, GL_RGBA, GL_RGBA, GL_UNSIGNED_BYTE);
-  delete converted_pxdata;
+  delete[] converted_pxdata;
   return ret;
 }
 
@@ -116,7 +116,7 @@ void graphics_push_texture_pixels(int texture, int width, int height, unsigned c
   glBindTexture(GL_TEXTURE_2D, get_texture_peer(texture));
   unsigned char* converted_pxdata = bgra2rgba(pxdata, width, height);
   glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA, width, height, 0, GL_RGBA, GL_UNSIGNED_BYTE, converted_pxdata);
-  delete converted_pxdata;
+  delete[] converted_pxdata;
 }
 
 } // namespace enigma
--- a/ENIGMAsystem/SHELL/Graphics_Systems/OpenGL-Common/textures.cpp
+++ b/ENIGMAsystem/SHELL/Graphics_Systems/OpenGL-Common/textures.cpp
@@ -29,6 +29,22 @@
 
 namespace enigma {
 
+// Because GLES won't do the conversion for us
+unsigned char* bgra2rgba(unsigned char* in, int w, int h) {
+  unsigned char* out = new unsigned char[w * h * 4];
+  int offset = 0;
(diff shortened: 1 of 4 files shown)


fundies:
android

RobertBColton:
Are you using GLAD or what? I remember what I think we did to solve the BGRA problem before, we used the extension which was supported on like 98% of hardware according to some Khronos statistics.
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_format_BGRA8888.txt

ENIGMAsystem/SHELL/Graphics_Systems/OpenGL-Common/textures.cpp
```diff
@@ -29,6 +29,22 @@
...
+      out[offset] = in[offset + 2];
+      out[offset + 1] = in[offset + 1];
+      out[offset + 2] = in[offset];
+      out[offset + 3] = in[offset + 3];
+      offset += 4;
+    }
+  }
+  return out;
+}
+
```
RobertBColton: Are you using GLAD or what? I remember what I think we did to solve the BGRA problem before, we used the extension which was supported on like 98% of hardware according to some Khronos statistics.
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_format_BGRA8888.txt

Edit: Correction* it's supported by 99.2% of devices.
https://opengles.gpuinfo.org/listextensions.php

BGRA GLES EXT Support

fundies: If i put bgra & bgra everything blue and red are swapped even on desktop. The real solution is to just kill this bgra nonsense in the engine but thats far too much work for me. Right now you have opengl doing the conversion but gles cannot do that conversion even if it supports bgra.

RobertBColton: No I mean put it back the way that it was (BGRA & **_RGBA_**) and add the extension. The desktop was obviously fine before you changed _anything_! I recall API tracing GMS and seeing that it was in fact still using BGRA internally and I also recall finding somewhere that it too used this extension.

RobertBColton: See GameMaker discussion as well:
https://forum.yoyogames.com/index.php?threads/switching-gms2-from-bgr-to-standard-rgb.33526/

fundies: The extension has nothing todo w/ conversion. bgra is already enable but GLES will not convert bgra to rgba regardless of any extension

RobertBColton: Try GL_BGRA_EXT

RobertBColton: Also TKG made a helper that appears to do the same thing:
https://github.com/enigma-dev/enigma-dev/pull/1904/files#diff-1246ecb7fd9dc46a9b17aa578b0d5c0bR37

JoshDreamland: I think Sam just checked this method in as part of his icon PR. Please try to reuse that.

RobertBColton: fundies is also being difficult and I don't know whether to believe him about BGRA.

RobertBColton:
Who's Travis?

fundies:
> Why's Travis failing?

Because I'm calling delete wrong somehow. I dunno what handles deletion of pxdata after uploading to gl. I haven't dug too deep yet

ENIGMAsystem/SHELL/Graphics_Systems/OpenGL-Common/textures.cpp
RobertBColton: Here's that thing tkg added.
https://github.com/enigma-dev/enigma-dev/blob/2d1220f23fa6dd057bf478eba8f25e5a82d34b6d/ENIGMAsystem/SHELL/Universal_System/image_formats.cpp#L188

PR #1912, opened 2020-03-10
[BGRA->ARGB] allow making prepending size optional relative to use case

time-killer-games:
XLib icon changing requires prepending the size (width, height) of the bitmap to the color grid image data as the first two values of the array. In most other cases, converting the colors of a bitmap will not want this information to be a part of that array, so I made an optional argument of a boolean to prepend the size to the dynamic array while keeping the default value for it to be `false`. This will help in the future to those who may need re-usability of that function to suit their needs more properly.

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/image_formats.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/image_formats.cpp
@@ -58,7 +58,7 @@ unsigned long *bgra_to_argb(unsigned char *bgra_data, unsigned pngwidth, unsigne
   unsigned char *bitmap = new unsigned char[bitmap_size]();
 
   unsigned i = 0;
-  unsigned elem_numb = pngwidth * pngheight + (prepend_size) ? 2 : 0;
+  unsigned elem_numb = pngwidth * pngheight + ((prepend_size) ? 2 : 0);
   unsigned long *result = new unsigned long[elem_numb]();
   if (prepend_size) {
     result[i++] = pngwidth; result[i++] = pngheight; // this is required for xlib icon hint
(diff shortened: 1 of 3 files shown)


time-killer-games:
@JoshDreamland I'm not sure whether adding one to widfull and hgtfull should be relative to the value of prepend_size: https://github.com/enigma-dev/enigma-dev/blob/e4545e17008fa1c4f1537242be568375d0b4c3fd/ENIGMAsystem/SHELL/Universal_System/image_formats.cpp#L56

Your insight is appreciated.

time-killer-games:

ENIGMAsystem/SHELL/Universal_System/image_formats.cpp
```diff
@@ -52,15 +52,17 @@ namespace enigma
...
-unsigned long *bgra_to_argb(unsigned char *bgra_data, unsigned pngwidth, unsigned pngheight) {
-  unsigned widfull = nlpo2dc(pngwidth) + 1, hgtfull = nlpo2dc(pngheight) + 1, ih,iw;
+unsigned long *bgra_to_argb(unsigned char *bgra_data, unsigned pngwidth, unsigned pngheight, bool prepend_size) {
+  unsigned widfull = nlpo2dc(pngwidth) + 1, hgtfull = nlpo2dc(pngheight) + 1, ih, iw;
   const int bitmap_size = widfull * hgtfull * 4;
   unsigned char *bitmap = new unsigned char[bitmap_size]();
 
   unsigned i = 0;
-  unsigned elem_numb = pngwidth * pngheight + 2;
+  unsigned elem_numb = pngwidth * pngheight + (prepend_size) ? 2 : 0;
```
JoshDreamland: This will always evaluate to 2 or 0, causing segmentation fault in the following code.

time-killer-games: either 2 or zero is the whole point, either i want to prepend the two extra values (width and height) or I don't... how would that cause a segmentation fault? Are you sure you looked at the entire code?

time-killer-games:
@JoshDreamland please mudge

PR #1913, opened 2020-03-10
fix texture leak

fundies:


fundies:
garbage directx. I'll fix that tmmrw. I wish @RobertBColton nvr wrote those systems tho

fundies:
@JoshDreamland dx fixed

PR #1915, opened 2020-03-11
Init window_get_icon_index() as -1

time-killer-games:
this is to prevent it from returning a valid sprite index when no sprite window icon was previously set


time-killer-games:
@JoshDreamland ready to be merged.

PR #1917, opened 2020-03-12
fixes emake output for "--list" & "--sever" and moves log files to /tmp

fundies:
Note: This brings back a lot of garbage printing during parse that should probably be manually hidden as it makes no sense to anyone but josh


fundies:
@JoshDreamland or @RobertBColton restart job that failed on appveyor and this should be ok for review/merge

PR #1925, opened 2020-03-20
[JDI] Remove Deprecated "register" Keyword

time-killer-games:


time-killer-games:
@JoshDreamland fundies wanted me to do this as a temporary measure until we can get JDI from your original repository with this problem fixed. Please merge when you can.

PR #1930, opened 2020-03-22
allow emake to specify enigma path

fundies:


CompilerSource/settings.h
```diff
@@ -94,7 +94,7 @@ struct CompilerInfo {
 };
 
 extern CompilerInfo compilerInfo;
-
+extern std::string enigma_root;
```
JoshDreamland: How confident are you that this will always be loaded before it's needed? What's our insurance policy?

Actually, seems this is part of our growing wad of static state. Can this be part of CompilerInfo or any of the other lumps of global state I've been trying to consolidate?

CommandLine/emake/OptionsParser.cpp
```diff
@@ -283,7 +290,7 @@ std::string OptionsParser::APIyaml(const buffers::resources::Settings* currentCo
 
 int OptionsParser::find_ey(const char* dir)
 {
-  fs::path targetDir(dir);
+  fs::path targetDir(_enigmaRoot + dir);
```
JoshDreamland:
```suggestion
  fs::path targetDir(_enigmaRoot);
  _enigmaRoot /= dir;
```

CompilerSource/settings-parse/parse_ide_settings.cpp
```diff
@@ -156,11 +164,16 @@ void parse_ide_settings(const char* eyaml)
...
   eobjs_directory = myReplace(escapeEnv(eobjs_directory), "\\","/");
   codegen_directory = myReplace(escapeEnv(codegen_directory), "\\","/");
+  enigma_root = myReplace(escapeEnv(enigma_root), "\\","/");
 #else
   eobjs_directory = escapeEnv(eobjs_directory);
   codegen_directory = escapeEnv(codegen_directory);
+  enigma_root = escapeEnv(enigma_root);
 #endif
 
+  if (enigma_root.back() != '/')
```
JoshDreamland: This stuff could probably also use to be fs::path, unless there's a C++17 chasm somewhere in the source tree.

CompilerSource/settings.h
fundies: well compiler info is stuff from your compiler.ey so I dont think it really fits there

fundies: parse_ide_settings or w/e should always be run before compile and that and compile are only things that need this var afaik

CompilerSource/settings-parse/parse_ide_settings.cpp
fundies: theres a lot to be redone here but I didnt want to do that yet

PR #1944, opened 2020-03-31
Freshen up newevents

fundies:
* merges master into newEvents
* merges egm refactor into newEvents
* fixes some build stuffs

Notes:
* running gm2egm currently gives error "EVENT ERROR: Someone is querying event data without parsing events."
* This does not fix the cmake build and only fixes the makefiles. I will address cmake after #1931 is merged


JoshDreamland:
Merged into NewEvents and EgmEvents.

PR #1948, opened 2020-04-02
turn down gles version so software renderer works

fundies:


JoshDreamland:
This will break shaders on some cards, right? But the game will continue to run even when they are bound?

fundies:
> This will break shaders on some cards, right? But the game will continue to run even when they are bound?

theres nothing gl3.2 specific in shader right now

PR #1949, opened 2020-04-02
[POSIX shell] these functions were written by a dummy

time-killer-games:
the fname argument needs quotes around it to accept spaces and it also needs a space between that and the args argument so the fname and first argument aren't fused together like Frankenstein's monster. Whoever wrote these functions initially was on crack. At least now, it works how it's supposed to, like it does on Windows, thanks to me. I can't believe this has been broken for the 10+ years ENIGMA has been around lol


time-killer-games:
@JoshDreamland idk why you haven't merged this yet but as I said on discord it is passing the yellow circle is a false status and it you click it it says all tests are finished and passing so please merge this when you can.

PR #1957, opened 2020-04-05
Fix SDL

time-killer-games:
this platform is useless and shouldn't exist, much like the person who wrote most of it.

I also intend to remove fundies from all copyrights in my personal fork, just for lolz.


time-killer-games:
@JoshDreamland i think this is the biggest improvement the project has seen in a very long time.

JoshDreamland:
You may disagree with decisions we've made. Try to do so in a constructive manner.

PR #1958, opened 2020-04-05
expand security measures to winblows

time-killer-games:


ENIGMAsystem/SHELL/Platforms/General/PFshell/PFshell.cpp
```diff
@@ -0,0 +1,46 @@
...
+namespace enigma_user {
+
+bool securemode_get_enabled() {
+  return secure_mode;
+}
+
+void securemode_set_enabled(bool enable) {
+  permission_requested = !enable;
+  secure_mode = enable;
+}
```
JoshDreamland, line 28: The purpose of this secure mode is *exactly* so the application doesn't just get to say "your computer is mine and I will now call out to any external programs I want." I am fine with creating some sort of environment variable for this so that developers don't have to okay their own program every time they go to run their game, but this lets developers okay themselves when *I* go to run their game.

ENIGMAsystem/SHELL/Widget_Systems/xlib/Info/About.ey
```diff
@@ -10,3 +10,6 @@ Author: Samuel Venable
 
 Depends:
   Build-Platforms: MacOSX, Linux, FreeBSD
+
+Represents:
+  Build-Platforms: Linux, FreeBSD
```
JoshDreamland, line 15: We should delay this until the vulnerability with `execute_program` is fixed. Otherwise arbitrary code execution (first-party or third-party) is as simple as showing a bad message.

ENIGMAsystem/SHELL/Platforms/General/PFshell/PFshell.cpp
time-killer-games: @JoshDreamland can you *please* not be this incredibly stupid? fundies is using you as a pawn of my frustration. No one will even know what this function is, it's not even documented and only people who are smart enough to know what the source code reads as will know, and those are the users who are competent enough to be trusted with such a feature.

ENIGMAsystem/SHELL/Widget_Systems/xlib/Info/About.ey
time-killer-games: @JoshDreamland what do you even mean? we already prompt the user whether they want to allow it. What more are you considering a vulnerability? I consider it a vulnerability that fundies as toxic of a person he is has any say in the projects direction and if find it crazy you see anything eye to eye with this total creep.

time-killer-games:
@JoshDreamland hey boi its been 10 hours and this hasn't been merged when i'm pretty sure you have nothing left to complain about.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -0,0 +1,107 @@
+#define byte __windows_byte_workaround
```
JoshDreamland, line 1: Please preserve the original copyright, here. If you changed more than a few lines, please add your name to it (looks like it's just the original, in which case, don't).

time-killer-games: @JoshDreamland i wrote execute_shell_for_output() on Windows, which is a lot of lines, and takes up more than half of the file. I also am the one who geot execute_program into a working state that doesn't crash by adding a proper waiting mechanism.

PR #1971, opened 2020-04-14
bkg tests

fundies:
* Adds some methods to check img data is equal in tests
* Adds some tests for background functions I plan to move to assArray next
* Adds some tests for sprite funcs I missed


ENIGMAsystem/SHELL/Universal_System/Resources/background.h
```diff
@@ -14,5 +14,10 @@ void background_set_alpha_from_background(int back, int copy_background, bool fr
 int background_get_texture(int backId);
 int background_get_width(int backId);
 int background_get_height(int backId);
-var sprite_get_uvs(int backId);
+var background_get_uvs(int backId);
+
+// Used for testing
+bool background_textures_equal(int id1, int id2);
+uint32_t background_get_pixel(int bkg, unsigned x, unsigned y);
```
JoshDreamland, line 21: Is this a GM method? GM used _getpixel. But these methods were *terribly* inefficient in GM, because they queried the GPU with each call. We're probably better off adding a unified pixbuf API. Well, we're probably way better off getting types working in ENIGMA properly so a pixbuf with a lifetime can be returned instead of an integer indicating an object without one. But I don't think it's wise to extend this old API.

fundies: this is just a method I added so I can check img data is correct / equal after various operations inside the test sogs

fundies: they aren't meant to be efficient and should only be used as means for testing. that's why I have the above comment

JoshDreamland: I see... you're only chcecking like ten pixels. Well, all right. Probably better to add some kind of region_has_color method, though, to avoid giving people bad ideas...

PR #1973, opened 2020-04-17
[Win32] Add function highscore_show(numb).

time-killer-games:
@RobertBColton This code is currently broken and I don't know how to get this fully working after more than 48 hours of writing and debugging it. I figured it's better to leave this pull request up for someone else to mess with than to trash all the work I've done. Hopefully someone with Windows experience, like Robert, or to some lesser extent, Josh and fundies, might be willing to finish this. All I know is, even Josh and fundies who rarely use Windows are probably much more likely to get this working than I am. You guys do what you want with this; I hope I didn't waste my time working on it.

Note the current highscore table that renders to the game window should be moved to Widgets = None, as suggested by fundies. If someone completes this I will be able to then copy the coding logic they used and adapt it to an XLib version, so we will have all our supported platforms covered. 

highscore_show


sharkwouter:
That looks pretty good, I'll test it tonight.

time-killer-games:
@sharkwouter Sorry for the confusion. I should've mentioned in the desciption this is a WIP.

time-killer-games:
@RobertBColton if you ever decide to finish this pr from where I left off, it's up to you, but if you do I have some documentation on what I've currently written provided below:

- The textbox control ID's on the left column of the grid are equal 1000 - 1009 from top to bottom.
- The textbox control ID's on the right column of the grid are equal 1010 - 1019 from top to bottom.
- If highscore_prev[previous_score_index] is less than highscore_array[current_score_index] then record new score on the list and everything scored less than or equal to the newly recorded score entry should be moved down by one textbox vertically on the two grid columns.
- I used the registry functions to record the highscore table values because this is what gamemaker 8.1 and below for Windows does. GameMaker for Mac 7.5 uses ini internally with the lack of a system registry, so that's what we'll do for Xlib, and I'll be the one to write the xlib version, copying the logic you might use in the windows version, but with xlib api's.

Issue #1980, opened 2020-04-18
Fullscreen displays wrong when building for SDL


sharkwouter:
Again I'm using 1945 as an example, it can be downloaded here: http://rupert.id.au/tutorials/gamemaker/examples.php

This game is configured to start on full screen by default. When using the platform Linux, this works. When using SDL on either Debian 10 or Window 10, it seems only a 4th of the room is shown in full screen mode. I've added a screenshot of both platform Linux and SDL for comparison.
![Screenshot from 2020-04-18 12-15-22](https://user-images.githubusercontent.com/5042659/79635026-96867000-816e-11ea-85ac-329822f16503.png)
![Screenshot from 2020-04-18 12-15-48](https://user-images.githubusercontent.com/5042659/79635027-971f0680-816e-11ea-96ce-43b2ce0cd79c.png)

sharkwouter:
To clarify, SDL full screen only shows a portion of the top left of the screen.

fundies:
Pretty sure tkg caused this disconnect between sdl & other platforms with his other fullscreen fixes but I haven't dug too deep on this yet

sharkwouter:
I don't think it matters who broke it tbh -.-

fundies:
it's not about who but when and why

fundies:
@JoshDreamland or @RobertBColton or @sharkwouter you can close this as was "fixed" in #1981

PR #1981, opened 2020-04-19
swap sdl full screen modes

fundies:
Swaps sdl's default fullscreen mode to borderless window to match win32/xlib. Scaling is still bugged in true full screen.

workaround for #1980


sharkwouter:
This seems to solve the issue for me.

PR #2021, opened 2020-05-21
Fix Fullscreen on Linux w/ Certain WM's

time-killer-games:
I don't know what WM Ubuntu Budgie has but one site said Mutter, so I guess that's the one.

Anyway, I discovered Ubuntu Budgie and several other distro's have WM's that don't like fullscreen switching when the window has a fixed size, and it will prevent the game from going fullscreen. This fixes the issue.

Weird, two fullscreen bugs I found, (and fixed), on two platforms, all in one night. Sheesh.

[reproducible.zip](https://github.com/enigma-dev/enigma-dev/files/4670554/reproducible.zip)

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
@@ -306,9 +306,10 @@ bool window_get_stayontop() {
   return windowHasAtom(a);
 }
 
+static bool set_sizeable = false;
 void window_set_sizeable(bool sizeable) {
   if (window_get_maximized()) return;
-  if (window_get_fullscreen()) return;
+  if (window_get_fullscreen() && !set_sizeable) return;
   enigma::isSizeable = sizeable;
   XSizeHints *sh = XAllocSizeHints();
   sh->flags = PMinSize | PMaxSize;
@@ -494,8 +495,9 @@ void window_set_position(int x, int y) {
   XMoveWindow(disp, win, x - wa.x, y - wa.y);
 }
 
+static bool set_size = false;
 void window_set_size(unsigned int w, unsigned int h) {
-  if (window_get_fullscreen()) return;
+  if (window_get_fullscreen() && !set_size) return;
   enigma::windowWidth = w;
   enigma::windowHeight = h;
   if (!enigma::isSizeable) {
@@ -519,6 +521,11 @@ void window_set_rectangle(int x, int y, int w, int h) {
 
 static bool prefer_sizeable = enigma::isSizeable;
 void window_set_fullscreen(bool full) {
+  if (window_get_fullscreen() != full && full) {
+    set_size = true;
+    window_set_size(enigma::windowWidth, enigma::windowHeight);
+    set_size = false;
+  }
   if (enigma::isFullScreen == full && !full) return;
   enigma::isFullScreen = full;
   if (full) {
@@ -541,8 +548,12 @@ void window_set_fullscreen(bool full) {
   xev.xclient.data.l[2] = 0;
   XSendEvent(disp, DefaultRootWindow(disp), False, SubstructureRedirectMask | SubstructureNotifyMask, &xev);
   if (!full) {
+    set_size = true;
     window_set_size(tmpSize::tmpW, tmpSize::tmpH);
+    set_size = false;
+    set_sizeable = true;
     window_set_sizeable(prefer_sizeable);
+    set_sizeable = false;
   }
 }
 
--- a/ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
@@ -567,15 +567,16 @@ void window_set_fullscreen(bool full) {
       window_set_rectangle(tmpX, tmpY, tmpWidth, tmpHeight);
     }
   } else {
-    prefer_sizeable = enigma::isSizeable;
     if (full) {
       tmpX = enigma::windowX;
       tmpY = enigma::windowY;
       tmpWidth = enigma::windowWidth;
       tmpHeight = enigma::windowHeight;
+      prefer_sizeable = enigma::isSizeable;
       window_set_fullscreen_helper(full);
     } else {
       window_set_fullscreen_helper(full);
+      prefer_sizeable = enigma::isSizeable;
       window_set_rectangle(tmpX, tmpY, tmpWidth, tmpHeight);
     }
   }
--- a/ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
@@ -517,15 +517,7 @@ void window_set_rectangle(int x, int y, int w, int h) {
 // FULLSCREEN //
 ////////////////
 
-namespace {
-
-bool prefer_sizeable = enigma::isSizeable;
-int tmpX = enigma::windowX;
-int tmpY = enigma::windowY;
-unsigned tmpWidth = enigma::windowWidth;
-unsigned tmpHeight = enigma::windowHeight;
-
-void window_set_fullscreen_helper(bool full) {
+void window_set_fullscreen(bool full) {
   if (enigma::isFullScreen == full && !full) return;
   enigma::isFullScreen = full;
   if (full) {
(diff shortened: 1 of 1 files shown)


time-killer-games:
I give up. Mutter is a fucking mess.

time-killer-games:
Nevermind, I figured it out.

time-killer-games:
@JoshDreamland hey josh, please merge this one when you get the chance, thanks!

time-killer-games:
@RobertBColton on Ubuntu Budgie, if you have the game start in fullscreen and switch to windowed mode, the window stays the same size as the entire screen, but simply gets window border with no size change. What I'm wondering, is how can I get the initial window size from roomsystem.cpp and use those values to be the initial size after switching to windowed mode for the first time. Do you have any idea how the initial window size is calculated in that file and if so, what variables in that file do I use to get the window dimensions? I've tried using enigma::windowWidth and enigma::windowHeight but it still gives me the full display size like can be seen in my code.

time-killer-games:
@JoshDreamland This is still a huge improvement over the previous behavior, but the only bug I can seem to fix that is left is if you have fullscreen set to true the in global game settings AND in your gml in the first room's (or object in the first room's) creation event, the window will always be sizeable when switching out of fullscreen. 

If you happen to start full screen using global game settings OR gml in that same creation event, (NOT _and_), then it will be sizeable relative to whether the user ticked the global game setting to make it sizeable when in windowed mode, which is the correct behavior.

Since people don't need to set the game fullscreen twice in a row using two different methods, this shouldn't be a huge problem. This should be ready for a merge when it passes.

ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
```diff
@@ -308,7 +308,7 @@ bool window_get_stayontop() {
 
 void window_set_sizeable(bool sizeable) {
   if (window_get_maximized()) return;
-  if (window_get_fullscreen()) return;
+  if (enigma::isFullScreen) return;
```
JoshDreamland, line 311: Is this part of the bug fix? This change implies that our internal representation of whether the game is fullscreen _will_ differ from the external representation, _**and**_ that when this happens, we should trust our own understanding over the WM's. Why is that the case? I would expect this situation to arise when the user tells their window manager to adjust the game's fullscreen state instead of telling the game to do so, and in that case, we would want to fix our internal state.

ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
```diff
@@ -517,13 +517,25 @@ void window_set_rectangle(int x, int y, int w, int h) {
 // FULLSCREEN //
 ////////////////
 
+static bool prefer_sizeable = false;
+static bool initial_fullscreen = true;
```
JoshDreamland, line 521: Please move these into the `window_set_fullscreen` method to avoid the temptation for other routines to misuse them.

ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
time-killer-games: It only differs during game window creation. Yes, this is part of the fix. enigma::isFullScreen is based on the global game setting. window_get_fullscreen() is based on state of the fullscreen atom. Since we initially start the game in a window, (and then make it fullscreen relative to the GGS) this is why it happens to work this way.

ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
time-killer-games: I would do that actually if it wouldn't break perfectly good working code, but it would....

time-killer-games: I misspoke; what you said seems correct.

time-killer-games: I tried exactly what you said and it broke the ability to set a sizeable window via global game settings. Reverted.

JoshDreamland: You dropped the word "static" when you moved them.

time-killer-games: Interesting, I though the static keyword would make the value constant like const does if put in a function. I'm not sure why I thought that... Anyway, that did the trick, thanks!

ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
RobertBColton: My question is why that's an issue, are you trying to call this in the create event? If so, maybe we need to rethink at what time we apply the GGS. Otherwise we could change the order that fullscreen and sizeable is applied from GGS so that fullscreen is applied first.

time-killer-games: @RobertBColton my guess Robert as to why this is an issue is due to the fact Ubuntu Budgie has a retarded WM. That's the only explanation I can think of. The issue this pr fixes doesn't happen with most WM's and platforms. Only reproducible in Ubuntu Budgie so far, but my guess is there are many other distro's using that WM.

RobertBColton: Why don't you just try moving the GGS fullscreen call above the sizeable call please.
https://github.com/enigma-dev/enigma-dev/blob/c7cb43b7db8d3943079d9433e7d37ef7b336a52e/ENIGMAsystem/SHELL/Universal_System/Resources/loading.cpp#L141

time-killer-games: I already know the technicalities as to why that would undo what i went out of my way to fix completely, but sure, i'll try it .

RobertBColton: No I mean, test it with this and on master.

time-killer-games: Pretty much what I was already doing.

RobertBColton: Well ok, try it then.

time-killer-games: I am giving you the benefit of a doubt, even though by default unless you tell the window manager you want a fixed sized window, it will be sizeable, which defeats the idea behind what you trying to explain, unless i misread something, which could be the case.

time-killer-games: Room set to 640 by 480 before and after switching out of fullscreen in master with this code in loading.cpp as i already predicted undid my efforts.

```
    // apply global game settings
    enigma_user::window_set_fullscreen(isFullScreen);
    enigma_user::window_set_sizeable(isSizeable);
    enigma_user::window_set_showborder(showBorder);
```
(dual monitor setup)

Before:
Screenshot from 2020-05-27 19-05-30

After:
Screenshot from 2020-05-27 19-05-35

time-killer-games: The code we have now is near perfection for all window managers. what's not to like?

time-killer-games: @RobertBColton I apologize if my posts above come across rude or snippy, that wasn't my intent. I'm trying to be more sensitive to these things.

RobertBColton: Maybe it would help you to look at the underlying code in SDL which deals with these issues.
https://fossies.org/linux/SDL2/src/video/x11/SDL_x11window.c

Take a look at specifically lines 925-960, 1259-1341, and 1345-1419.

RobertBColton: And also it would be useful to know if SDL reproduces these issues on Ubuntu Budgie or not.

time-killer-games: ` 1279             if (fullscreen) {
 1280                 /* we are going fullscreen so turn the flags off */
 1281                 sizehints->flags &= ~(PMinSize | PMaxSize);
 1282             } else {
 1283                 /* Reset the min/max width height to make the window non-resizable again */
 1284                 sizehints->flags |= PMinSize | PMaxSize;
 1285                 sizehints->min_width = sizehints->max_width = window->windowed.w;
 1286                 sizehints->min_height = sizehints->max_height = window->windowed.h;
 1287             }`

They appear to be doing the same exact thing I am here, only they are removing the min/max size hints to allow fullscreen rather than making the window sizeable the other way, which is set the min/max to INT_MAX. When they switch out of fullscreen, they restore the window to the size it previously was recorded to be before going fullscreen, which I also do in this pr.

So the educated guess would be, since they are doing the same thing effectively, just a very slightly different means to accomplish the same desired outcome, SDL doesn't have this issue, although I'll go ahead and test it, I seriously doubt they fucked it up like ISM did.

time-killer-games: @RobertBColton i uninstalled Ubuntu Budgie last night because I got tired of using it, are you going to require i install it again and test with SDL before this will get merged? I will if i have to. I noticed some regressions with mg_keystroke's game, but I'm not sure if this pr causes it at all, I'm going to check out master again and see whether it's a new regression in this pr.

time-killer-games:
I said this information to Josh already but i figured it might be good for you to see as well @RobertBColton 

Screenshot from 2020-05-29 17-01-13

so in my pr there is a regression with mg keystroke's game

Screenshot from 2020-05-29 17-53-51

when you get past the title screen

i was wondering whether it was really a regression my pr introduced 

or

if it was something wrong with enigma that already existed that my pr simply uncovered
that's literally the only game i can throw at my pr that has dumb problems

but it fixes fullscreen for all games including this one

it just has a side effect reproducible in 1 stinkin game

RobertBColton:
That looks like scaling or the viewport clip in the second one. As if it wasn't able to fully clear the whole screen in time.

time-killer-games:
@RobertBColton @JoshDreamland what do you guys suggest I do? I tried an imitation of SDL's code and it cause the same issues with that game. I don't want that game to break, but i also want reliably working fullscreen. So I feel pretty stuck here.

time-killer-games:
I changed one file with 19 lines added and one removed. If you pretend that needs breaking up into different pr's you clearly are trolling me.

PR #2029, opened 2020-06-19
Fix Execute Shell for Output Freeze

time-killer-games:
fixes a problem where execute_shell_for_output would hang the executed program whenever the output buffer exceeds 4096 characters.

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -63,6 +63,7 @@ void execute_shell(string fname, string args) {
 }
 
 string execute_shell_for_output(const string &command) {
+  string output;
   tstring tstr_command = widen(command);
   wchar_t ctstr_command[32768];
   wcsncpy(ctstr_command, tstr_command.c_str(), 32768);
@@ -108,7 +109,7 @@ string execute_shell_for_output(const string &command) {
     CloseHandle(hStdInPipeWrite);
     CloseHandle(pi.hProcess);
     CloseHandle(pi.hThread);
-    return buffer;
+    return output;
   }
   return "";
 }
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -77,17 +77,17 @@ string execute_shell_for_output(const string &command) {
   if (ok == FALSE) return "";
   ok = CreatePipe(&hStdOutPipeRead, &hStdOutPipeWrite, &sa, 0);
   if (ok == FALSE) return "";
-  STARTUPINFOW si = { };
+  STARTUPINFOW si = { 0 };
   si.cb = sizeof(STARTUPINFOW);
   si.dwFlags = STARTF_USESTDHANDLES;
   si.hStdError = hStdOutPipeWrite;
   si.hStdOutput = hStdOutPipeWrite;
   si.hStdInput = hStdInPipeRead;
-  PROCESS_INFORMATION pi = { };
-  if (CreateProcessW(NULL, ctstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
+  PROCESS_INFORMATION pi = { 0 };
+  DWORD dwRead = 0;
+  DWORD dwAvail = 0;
+  if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
     while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
-      DWORD dwRead = 0;
-      DWORD dwAvail = 0;
       PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
       if (dwAvail) {
         vector<char> buffer(dwAvail);
@@ -101,6 +101,15 @@ string execute_shell_for_output(const string &command) {
       if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
         TranslateMessage(&msg);
         DispatchMessage(&msg);
+      }
+    }
+    PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
+    if (dwAvail) {
+      vector<char> buffer(dwAvail);
+      BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
+      if (success || dwRead) {
+        buffer[dwAvail] = 0;
+        output.append(buffer.data(), dwAvail);
       }
     }
     CloseHandle(hStdOutPipeWrite);
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -89,8 +89,8 @@ string execute_shell_for_output(const string &command) {
     CloseHandle(pi.hThread);
     CloseHandle(hStdOutPipeWrite);
     CloseHandle(hStdInPipeRead);
-    char buffer[BUFSIZ];
     for (;;) {
+      char buffer[BUFSIZ];
       DWORD dwRead = 0;
       BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
       if (success || dwRead) {
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -87,37 +87,23 @@ string execute_shell_for_output(const string &command) {
   DWORD dwRead = 0;
   DWORD dwAvail = 0;
   if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
-    while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
-      PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
-      if (dwAvail) {
-        vector<char> buffer(dwAvail);
-        BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
-        if (success || dwRead) {
-          buffer[dwAvail] = 0;
-          output.append(buffer.data(), dwAvail);
-        }
-      }
-      MSG msg;
-      if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
-        TranslateMessage(&msg);
-        DispatchMessage(&msg);
-      }
-    }
-    PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
-    if (dwAvail) {
-      vector<char> buffer(dwAvail);
-      BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
-      if (success || dwRead) {
-        buffer[dwAvail] = 0;
-        output.append(buffer.data(), dwAvail);
-      }
-    }
-    CloseHandle(hStdOutPipeWrite);
-    CloseHandle(hStdInPipeRead);
-    CloseHandle(hStdOutPipeRead);
-    CloseHandle(hStdInPipeWrite);
     CloseHandle(pi.hProcess);
     CloseHandle(pi.hThread);
+    CloseHandle(hStdOutPipeWrite);
+    CloseHandle(hStdInPipeRead);
+    char buffer[BUFSIZ];
+    BOOL success = TRUE;
+    DWORD dwRead = 0;
+    for (;;) {
+      BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
+      if (success || dwRead) {
+        buffer[dwRead] = 0;
+        output.append(buffer, dwRead);
+        buffer[0] = 0;
+      } else { break; }
+    }
+    CloseHandle(hStdOutPipeRead);
+    CloseHandle(hStdInPipeWrite);
     return output;
   }
   return "";
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -84,8 +84,6 @@ string execute_shell_for_output(const string &command) {
   si.hStdOutput = hStdOutPipeWrite;
   si.hStdInput = hStdInPipeRead;
   PROCESS_INFORMATION pi = { 0 };
-  DWORD dwRead = 0;
-  DWORD dwAvail = 0;
   if (CreateProcessW(NULL, ctstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
     CloseHandle(pi.hProcess);
     CloseHandle(pi.hThread);
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -86,21 +86,21 @@ string execute_shell_for_output(const string &command) {
   PROCESS_INFORMATION pi = { };
   if (CreateProcessW(NULL, ctstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
     while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
-      MSG msg;
-      if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
-        TranslateMessage(&msg);
-        DispatchMessage(&msg);
-      }
       DWORD dwRead = 0;
       DWORD dwAvail = 0;
       PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
       if (dwAvail) {
-        char buffer[dwAvail];
-        BOOL success = ReadFile(hStdOutPipeRead, buffer, dwAvail, &dwRead, NULL);
+        vector<char> buffer(dwAvail);
+        BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
         if (success || dwRead) {
           buffer[dwAvail] = 0;
-          output.append(buffer, dwAvail);
+          output.append(buffer.data(), dwAvail);
         }
+      }
+      MSG msg;
+      if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
+        TranslateMessage(&msg);
+        DispatchMessage(&msg);
       }
     }
     CloseHandle(hStdOutPipeWrite);
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -86,7 +86,7 @@ string execute_shell_for_output(const string &command) {
   PROCESS_INFORMATION pi = { 0 };
   DWORD dwRead = 0;
   DWORD dwAvail = 0;
-  if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
+  if (CreateProcessW(NULL, ctstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
     CloseHandle(pi.hProcess);
     CloseHandle(pi.hThread);
     CloseHandle(hStdOutPipeWrite);
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -90,9 +90,8 @@ string execute_shell_for_output(const string &command) {
     CloseHandle(hStdOutPipeWrite);
     CloseHandle(hStdInPipeRead);
     char buffer[BUFSIZ];
-    BOOL success = TRUE;
-    DWORD dwRead = 0;
     for (;;) {
+      DWORD dwRead = 0;
       BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
       if (success || dwRead) {
         buffer[dwRead] = 0;
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -99,7 +99,6 @@ string execute_shell_for_output(const string &command) {
       if (success || dwRead) {
         buffer[dwRead] = 0;
         output.append(buffer, dwRead);
-        buffer[0] = 0;
       } else { break; }
     }
     CloseHandle(hStdOutPipeRead);
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
@@ -90,21 +90,25 @@ string execute_shell_for_output(const string &command) {
         TranslateMessage(&msg);
         DispatchMessage(&msg);
       }
+      DWORD dwRead = 0;
+      DWORD dwAvail = 0;
+      PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
+      if (dwAvail) {
+        char buffer[dwAvail];
+        BOOL success = ReadFile(hStdOutPipeRead, buffer, dwAvail, &dwRead, NULL);
+        if (success || dwRead) {
+          buffer[dwAvail] = 0;
+          output.append(buffer, dwAvail);
+        }
+      }
     }
     CloseHandle(hStdOutPipeWrite);
     CloseHandle(hStdInPipeRead);
-    char buffer[4096] = { };
-    DWORD dwRead = 0;
-    ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
-    while (ok == TRUE) {
-      buffer[dwRead] = 0;
-      ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
-    }
     CloseHandle(hStdOutPipeRead);
     CloseHandle(hStdInPipeWrite);
     CloseHandle(pi.hProcess);
     CloseHandle(pi.hThread);
-    return shorten(widen(buffer));
+    return buffer;
   }
   return "";
 }


time-killer-games:
@RobertBColton ready for review when you are, thanks!

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -90,21 +91,25 @@ string execute_shell_for_output(const string &command) {
...
     }
     CloseHandle(hStdOutPipeWrite);
     CloseHandle(hStdInPipeRead);
-    char buffer[4096] = { };
-    DWORD dwRead = 0;
-    ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
-    while (ok == TRUE) {
-      buffer[dwRead] = 0;
-      ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
-    }
```
RobertBColton, line 102: Why was this reading twice?

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -90,21 +91,25 @@ string execute_shell_for_output(const string &command) {
...
     CloseHandle(hStdOutPipeRead);
     CloseHandle(hStdInPipeWrite);
     CloseHandle(pi.hProcess);
     CloseHandle(pi.hThread);
-    return shorten(widen(buffer));
+    return output;
```
RobertBColton, line 112: What makes you think output is UTF-8 from this API?

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -90,21 +91,25 @@ string execute_shell_for_output(const string &command) {
         TranslateMessage(&msg);
         DispatchMessage(&msg);
       }
+      DWORD dwRead = 0;
+      DWORD dwAvail = 0;
+      PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
+      if (dwAvail) {
+        char buffer[dwAvail];
```
RobertBColton, line 98: This is a variable-length array, VLAs are not supported in standard C++. You would need to call new here or use a fixed size buffer and read in a loop, as I noted in the below comment.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
time-killer-games: That's kinda what I am doing now. That's why I changed it.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
time-killer-games: Josh told me calling shorten(widen(something)) returns nothing different from what something already would return by itself, and that if my code worked with utf8 (which it does in my testing) that would imply i can just return the buffer appended to the output variable directly instead of calling shorten and widen on it, which has proven to be redundant, as Josh suspected.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
time-killer-games: fundies hates C's `new` keyword so I will honor his wishes and use std::vector. That makes it easier, anyway...

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: Josh told me the synchronous version of the API shouldn't have needed a loop, so I still don't get why it was reading twice.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: C does not have a `new` operator, C++ does, C has malloc.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
time-killer-games: The process is garanteed to hang once it gets a certain number of bytes to the buffer, if you try to do it after the process has already quit. It needs to be done during the time it is being run, otherwise the hanging will occur. I am not the only one who has run into this problem 

> The problem is that windows buffers just a limited amount of data. So you must read the pipe while the process is still running, otherwise the process will be blocked because it cannot write any more data to the pipe.

https://stackoverflow.com/a/13894991/4821390

time-killer-games: See also - https://forum.yoyogames.com/index.php?threads/execute-shell-for-windows-macos-and-ubuntu.7145/post-452964

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
time-killer-games: Well all i know is he doesn't like C approaches and he doesn't like new so i thought the reason he didn't like new was because it was tied to c. apparently not.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
time-killer-games: @RobertBColton Alright, I think I addressed the issue. I'm going to test it real quick before I conclude we've made progress. Just want to make sure I did it right.

time-killer-games: @RobertBColton alright, I tested and it's working ok

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: The vector is neater because then you don't have to `delete` it.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -85,6 +86,17 @@ string execute_shell_for_output(const string &command) {
   PROCESS_INFORMATION pi = { };
   if (CreateProcessW(NULL, ctstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
     while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
+      DWORD dwRead = 0;
+      DWORD dwAvail = 0;
+      PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
+      if (dwAvail) {
+        vector<char> buffer(dwAvail);
+        BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
```
RobertBColton, line 94: Just a note, you don't have to do this, but `&buffer[0]` is also the same as `buffer.data()` which you could arguably say is cleaner.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: I understand now what happened, the reading twice is what was causing the freezing because it was blocking waiting to read more bytes than there were available, it's very obvious. Regardless, your solution of obtaining the available bytes is much easier than implementing this as a loop, excellent work.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: Josh explained to me now as well that is does not make a difference. It is not standardized what format stdout happens to be in. This lies at the "boundary" of the program and therefore is not something we need to worry about.
http://utf8everywhere.org/

We could perhaps later provide shorten and widen in the EDL user land though. For now, there's nothing else to be said about this.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: I see, the hang was the result of attempting to read more than the program would ever output.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -85,6 +86,17 @@ string execute_shell_for_output(const string &command) {
   PROCESS_INFORMATION pi = { };
   if (CreateProcessW(NULL, ctstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
     while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
```
RobertBColton, line 88: One more thing to do. I want you to try `WaitForSingleObject(hStdOutPipeRead, 5)` instead of waiting on the process itself which actually wakes up more often than necessary.

time-killer-games: Yes, I can try that. Although, I noticed a new problem, which what you suggested may or may not fix it. With programs that open and close immediately they will sometimes produce output, other times an empty string, even though in both cases we are using the same code. An example of this can be cmd /c echo poop. Sometimes it will return poop, other times an empty string. I think I might know why that is if you idea doesn't work, but I'll try what you said first.

time-killer-games: Ok, I tried this, and it returned an empty string and nothing else, every time. It's weird, because it really doesn't makes sense to me why it is doing that

time-killer-games: Ok, I figured it out, the problem with it returning an empty string was solved by reading output after the program closed, and it does not hang the command past 4096 characters i verified from testing and that's apparently because I am also reading output while it is still running.

RobertBColton: Undo your last commit, I'm going to tell you how to fix it.
https://docs.microsoft.com/en-us/windows/win32/procthread/creating-a-child-process-with-redirected-input-and-output?redirectedfrom=MSDN

RobertBColton: Basically, you need to read in a loop here like in that link, but _without_ waiting for any object. You should use the success return of `ReadFile` to know when to terminate the loop. You should not need to do _anything_ after the process has exited like you're trying to do at the moment either.

time-killer-games: @RobertBColton why should i change perfectly working code? I tested exactly what you said and it broke everything. It went back to returning an empty string every time. I'm not going to revert these changes until you can find a replacement that actually does what we need. But I see no need to replace it to begin with, since it does what it needs to right now.


This is exactly what I tried, and I got an empty string:
```
    for (;;) {
      PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
      if (dwAvail) {
        vector<char> buffer(dwAvail);
        success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
        if (success && dwRead) {
          buffer[dwAvail] = 0;
          output.append(buffer.data(), dwAvail);
        } else { break; }
      } else { break; }
    }
```

RobertBColton: Try again without using PeekNamedPipe, the Microsoft tutorial I linked does not use that, it's the source of the problem.

time-killer-games: Ok, I'll give it a shot, I wish I knew that in the first place before I got so cocky lel

time-killer-games: Ok so i tried this and it hangs.

```
    BOOL success = FALSE;
    char buffer[BUFSIZ];
    DWORD dwRead = 0;
    for (;;) {
      success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
      if (success || dwRead) {
        buffer[dwRead] = 0;
        output.append(buffer, dwRead);
      } else { break; }
    }
```

time-killer-games: Any chance we could just use my working code instead of trying to solve a riddle that might not even exist?

RobertBColton: Change ` buffer[dwRead] = 0;` to `buffer[0] = 0;`

time-killer-games: Thank you for babysitting me

RobertBColton: It might need to be `buffer[0] = '\0';` try that too.

RobertBColton: Also you need to append it to the string first! Omg, how did you not see that before.

time-killer-games: this: `buffer[0] = '\0';` hangs
this: `buffer[0] = 0;` echos zero even though i wasn't echoing zero.

time-killer-games: What I'm actually echoing:
```
var str = "";
repeat (2046) { str += "10"; }
show_message(echo(str));
```
where as echo is a helper script function that does this:
```
var command = "";
if (os_type == os_windows) 
command = "cmd /c @echo off & echo " + string_replace_all(argument0, "&", "^&");
else command = "echo " + string(argument0);
return execute_shell_for_output(command);
```

time-killer-games: @RobertBColton deleting my comments when you falsely interpret them as rude doesn't accomplish anything whatsoever

RobertBColton: Ok try this:
```cpp
      if (success || dwRead) {
        buffer[dwRead] = 0;
        output.append(buffer, dwRead);
        buffer[0] = 0;
      } else { break; }
```

RobertBColton: Also try just `if (success)` because I am not sure if it's possible for dwRead to be non-zero while success is false.

RobertBColton: @time-killer-games calm down, I deleted mine too because I accidentally gave you erroneous suggestions.

time-killer-games: lol ok, I actually was parodying something fundies told me. Don't take it seriously.

Trying it now.

RobertBColton: TKG, if you think about it, it doesn't make sense why the Microsoft recommended solution would hang. I believe there's something subtle being missed either in the creation of the pipe or process.

time-killer-games: We aren't trying to achieve the same goals as Microsoft's example is, for one thing. They are doing quite a number of things in that code example we aren't doing, so trying to adapt existing code for completely different purpose doesn't seem like the best way to go about it.

time-killer-games: For reference, this user did the same exact thing and it worked for them for all the same reasons it worked for me, that reason is it happens to be one of many working solutions: https://stackoverflow.com/a/13895520/4821390

RobertBColton: The extra things in the Microsoft example do not affect the way it reads data from the child process in any way though.

time-killer-games: Now I realize, it could be a bug with gamemaker itself that could be the culprit, there are quite a number of extensions i wrote for GM that ended up not working in GM but did work in ENIGMA. If you want, we can try it again but this time do it purely in ENIGMA, but I do have some doubts it'll make a difference.

```
var str = "";
repeat (2046) { str += "10"; }
show_message(echo(str));
```
Capture

time-killer-games: PS i did try it with above 2048 in the repeat; that code was used because i wanted the 1's and 0's to form a perfect square  lel

so before I get any further, do want me to try your last suggestion? Or do you have any modifications on it you would like to make first?

RobertBColton: No, that suggestion sucked, I have a new one. I'm pretty sure I have the solution now.

RobertBColton: As it says in the Microsoft tutorial again. I want you to move the code which closes the handles to right after `CreateProcess` and the read loop should come AFTER the file handles are already closed. It may sound counter intuitive, but I see other people saying it and it is the main difference between your version here and Microsoft's.

time-killer-games: Fascinating. Try it now.

RobertBColton: https://stackoverflow.com/a/10857336

time-killer-games:
```
  if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
    while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
      MSG msg;
      if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
      }
    }
    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);
    CloseHandle(hStdOutPipeWrite);
    CloseHandle(hStdInPipeRead);
    CloseHandle(hStdOutPipeRead);
    CloseHandle(hStdInPipeWrite);
    DWORD dwRead = 0;
    DWORD dwAvail = 0;
    PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
    if (dwAvail) {
      vector<char> buffer(dwAvail);
      BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
      if (success || dwRead) {
        buffer[dwAvail] = 0;
        output.append(buffer.data(), dwAvail);
      }
    }
```

Judging by that red box, i messed up the master plan.

RobertBColton: Yes, looks like you maybe did, also you need to be using the other for loop you wrote, not that old one you have there. We're not using PeekNamedPipe or WaitForSingleObject, and neither does the Microsoft example.

RobertBColton: Also, I have a secret to tell you about shark poop.

time-killer-games: Yay! You did it robert! Super awesome gay bitch 999999x!!!!

```
  if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
    while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
      MSG msg;
      if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
      }
    }
    pid = pi.dwProcessId;
    prevpid = pid;
    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);
    DWORD dwRead = 0;
    DWORD dwAvail = 0;
    PeekNamedPipe(hStdOutPipeRead, NULL, 0, NULL, &dwAvail, NULL);
    if (dwAvail) {
      vector<char> buffer(dwAvail);
      BOOL success = ReadFile(hStdOutPipeRead, &buffer[0], dwAvail, &dwRead, NULL);
      if (success || dwRead) {
        buffer[dwAvail] = 0;
        output.append(buffer.data(), dwAvail);
      }
    }
    CloseHandle(hStdOutPipeWrite);
    CloseHandle(hStdInPipeRead);
    CloseHandle(hStdOutPipeRead);
    CloseHandle(hStdInPipeWrite);
```

Capture

RobertBColton: Good! Now get rid of the PeekNamedPipe or I'm not telling you about the shark poop!

time-killer-games: PeekNamedPipe gets rid of the need for a second/multiple loops though? It gets the correct buffer size all in one if statement, as apposed to multiple loop checks. I can do it if you want, but could explain the science behind it?

time-killer-games: Nevermind i see it now, it's because I set it to 4096 again and it started hanging.

time-killer-games: what do you want me to do in place of that? I want to know what to do and do it right the first time

RobertBColton: The `for(;;)` loop you tried earlier from that same Microsoft example!

time-killer-games: ok cool!

time-killer-games: don't leave my titties hanging!

```
  if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
    while (WaitForSingleObject(pi.hProcess, 5) == WAIT_TIMEOUT) {
      MSG msg;
      if (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
      }
    }
    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);
    char buffer[BUFSIZ];
    BOOL success = TRUE;
    DWORD dwRead = 0;
    for (;;) {
      BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
      if (success || dwRead) {
        buffer[dwRead] = 0;
        output.append(buffer, dwRead);
        buffer[0] = 0;
      } else { break; }
    }
    CloseHandle(hStdOutPipeWrite);
    CloseHandle(hStdInPipeRead);
    CloseHandle(hStdOutPipeRead);
    CloseHandle(hStdInPipeWrite);
```

time-killer-games: I'm getting rid of that waitforsingleobject now i forgot about that

RobertBColton: Did it work or not? if it worked, are you going to actually commit it and fix the PR?

time-killer-games: my tits are still hanging
```
  if (CreateProcessW(NULL, cwstr_command, NULL, NULL, TRUE, CREATE_NO_WINDOW, NULL, NULL, &si, &pi)) {
    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);
    char buffer[BUFSIZ];
    BOOL success = TRUE;
    DWORD dwRead = 0;
    for (;;) {
      BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
      if (success || dwRead) {
        buffer[dwRead] = 0;
        output.append(buffer, dwRead);
        buffer[0] = 0;
      } else { break; }
    }
    CloseHandle(hStdOutPipeWrite);
    CloseHandle(hStdInPipeRead);
    CloseHandle(hStdOutPipeRead);
    CloseHandle(hStdInPipeWrite);
```

time-killer-games: tits are hanging = program is hanging

RobertBColton: Also accordingly you are still closing the wrong handles! You need to close the other two bottom handles.

Handles to Close After CreateProcess

RobertBColton: @time-killer-games You're almost there, don't give up now!

RobertBColton: WaitForTKG();

time-killer-games: Thanks Robert joos one SEXY bitch! It werk!

```
    CloseHandle(pi.hProcess);
    CloseHandle(pi.hThread);
    CloseHandle(hStdOutPipeWrite);
    CloseHandle(hStdInPipeRead);
    char buffer[BUFSIZ];
    BOOL success = TRUE;
    DWORD dwRead = 0;
    for (;;) {
      BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
      if (success || dwRead) {
        buffer[dwRead] = 0;
        output.append(buffer, dwRead);
        buffer[0] = 0;
      } else { break; }
    }
    CloseHandle(hStdOutPipeRead);
    CloseHandle(hStdInPipeWrite);
```

RobertBColton: Ok, now commit so we can get this merged. Also, leave out the WaitForSingleObject message pump, it doesn't actually work. You can also get rid of the single line `buffer[0] = 0;` I told you to add, it's not needed.

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -76,35 +77,34 @@ string execute_shell_for_output(const string &command) {
...
-    ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
-    while (ok == TRUE) {
-      buffer[dwRead] = 0;
-      ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
+    for (;;) {
+      BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
+      if (success || dwRead) {
+        buffer[dwRead] = 0;
+        output.append(buffer, dwRead);
+        buffer[0] = 0;
```
RobertBColton, line 102: Delete this single line I had you add, it's not needed now.

time-killer-games: Yee

time-killer-games:
@RobertBColton I LOVE YOU, BITCH!

RobertBColton:
Anyway, as the CI builds, I was getting at shark poop, it comes in various shapes and sizes, but you can never quite predict how long it will truly last.

time-killer-games:
@RobertBColton Unfortunately, it crashes now. If you try to move do anything with the game window while a program is still running. :/ my previous code didn't have this problem because of the sleep interval WaitForSingleObject has

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
```diff
@@ -76,35 +77,33 @@ string execute_shell_for_output(const string &command) {
...
-    ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
-    while (ok == TRUE) {
-      buffer[dwRead] = 0;
-      ok = ReadFile(hStdOutPipeRead, buffer, 4095, &dwRead, NULL);
+    for (;;) {
+      BOOL success = ReadFile(hStdOutPipeRead, buffer, BUFSIZ, &dwRead, NULL);
+      if (success || dwRead) {
+        buffer[dwRead] = 0;
+        output.append(buffer, dwRead);
+      } else { break; }
```
RobertBColton, line 102: These lines could have been written more simply, but I'm not going to nitpick as it's not a big deal.
```cpp
if (!success || !dwRead) break;
buffer[dwRead] = 0;
output.append(buffer, dwRead);
```

time-killer-games: Did you read what I said and that it crashes now?

RobertBColton:
Ok, so here's the skinny.

RobertBColton:
It's not crashing, I presume you are referring to it hanging then Windows fading the window out. This is the result of the ReadFile performing a non pumping wait on the pipe.

RobertBColton:
Can you tell me whether the regular execute_shell is a **_blocking_** function? Say you execute a long running task with it, does GM freeze or does the game keep playing?

ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSshell.cpp
RobertBColton: Yes, answer my questions, I already know how to fix it. The fix is not necessarily simple though.

time-killer-games:
Regular execute shell is not blocking. But my process information extension basically allows for it to give the option to either block or not block the main thread. 

The reason execute shell for output blocks is so we can wait for output to finish printing so it may be returned all at once.

RobertBColton:
Ok, long story short, I'm merging this as it is right now and we can fix the issue in a follow up.

There are two ways to fix it but the basic answer is MsgWaitForMultipleObjects with the asynchronous I/O API which allows waiting for both messages and events. We can do that to produce a message-pump-read loop similar to the following.
```cpp
while(Wait for Message || Wait for Data) {
  PumpMessages();
  ReadFileAndAppendOutput();
}
```

The alternative would be putting the read loop you have in a thread, which I had because then we have to deal with thread synchronization.

time-killer-games:
@RobertBColton what's the point of merging it now? Can't I just fix it first?

RobertBColton:
I'm preferring not to because the review chain got too big here, and I expect it to take us a bit of discussion to actually achieve the above loop. I'd rather do that in a nice clean space on a new pull request.

RobertBColton:
Letting me know if #2030 actually worked is the quickest way to speed this up.

time-killer-games:
That's fine.

RobertBColton:
It's up to you, this was your pull request though, so he couldn't have commented on it. If I make it then he can comment on it. Now that I've merged this, all you need to do is take what I did here and integrate it with #2030 to create the while loop I described above.

RobertBColton:
The loop will look something like the following.

```cpp
    HANDLE waitHandles[] = { pi.hProcess, };
    while (DWORD eventSignalId = MsgWaitForMultipleObjects(2, waitHandles, false, INFINITE, QS_ALLEVENTS)) {
      if (eventSignalId == WAIT_OBJECT_0) break;
      MSG msg;
      while (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
      }
      ReadFileAsync();
      AppendOutput();
    }
```

RobertBColton:
@time-killer-games one more thing, as a temporary work around for your own personal games/extensions, you can add the simple message pump for the window back until you figure out named pipes. This may work better than before, but it may still have the exact same 4096 hang problem as before, because it's limited by the amount of output Windows can hold in its internal buffer. Just a recommendation if you want to get a fix out to people quickly.

RobertBColton:
Also feel free to let me know how it goes too, because I am very curious about this Windows buffer size or if in fact the original problem was just you not having those other two handles closed.

PR #2076, opened 2020-07-07
ignore

fundies:


time-killer-games:
3) things:

1) I ran into the exact same problem with the compilers on BSD, Robert or somebody needs to fix whatever part of lateralGM and/or the plugin that forces us to rely on the existence of gcc.ey, and not being able to just use clang.ey instead if gcc.ey is not found. Something is screwed up in such a way that you cannot simply make gcc.ey reference clang internally and call it done. It's not that simple because whoever wrote the non-engine parts of the project hardcoded it so that we need gcc.ey to exist AND it actually needs to reference gcc internally.

 2) due to 1) the temporary fix would be to keep gcc.ey and clang.ey untouched, other than perhaps making clang the default compiler if it isn't already in master. Don't delete or modify either one other earn that, and then this pr will mostly be ready for getting Mac to work again.

3) lastly, oddly enough, I'm getting a "uname": no such file or directory error at lateralGM startup. Because it couldn't find or launch uname, we ended up with it searching for DLL's instead of having the actual DYLIB extension the DYLIB's we compiled actually have. This can be fixed by adding the full path to uname i.e. shell /usr/bin/uname under Config.mk and voila!

fundies:
gcc is just alias to clang for me. This pr mostly works but theres a segfault in jdi that josh didn't want to help with so its on hold

PR #2089, opened 2020-07-25
EGM Events

JoshDreamland:
Reworks the entirety of the event system, redoing some or all logic for event parsing, event generation, event triggering, inheritance, etc.

This brings a few key advantages:
1. **More event types.** We are no longer bound by events supported by LateralGM, or any other toolkit.
2. **Dynamic event description.** It allows us name events whatever we like, and parameterize them with as many arguments as we need.
    * Joystick events can now accept a joystick number parameter AND a button/axis number.
    * Alarm events are no longer limited in number to the 16 or whatever that Mark squeezed in.
    * We don't have to keep making up strange values for new "special" events.
3. **EGM Support.** Dynamic naming is part of this retooling. This means filenames in EGM can be `Collision[obj_wall].edl`. This unblocks EGM support.

This comes with great risk:
1. **Build errors.** Some variable or event declarations may be emitted twice.
2. **Link errors.** Some event declarations may not have their body emitted, leading to a link error.
3. **Behavioral regressions.** There may be quirks of object inheritance we don't currently have tests for. We may not notice these during manual QA testing.

We will test this on our CI and on various user games before merging. This is not a guarantee, but should mitigate noticeable regressions. If you have a game to test these changes against, please clone this branch and test them when the CI is green.

## Details
The topology of parsed entity storage has been ***greatly*** reworked.
* Parameterized event types are now more specific.
    * `SYSTEM` is no longer an event type specification. It has been replaced by two others:
        1. `TriggerOnce` if no iteration support is required (e.g. Create events),
        2. `TriggerAll` if a list of them is to be maintained for iteration.
    * `SPECIAL` is now `SPECIALIZED`, and actually works for parameterized events.
        * Prior to this refactor, `SPECIAL` just meant that different values of an event's ID meant entirely different behaviors (look at `ev_other`: `ev_outside` and `ev_game_start` have nothing in common).
        * After this refactor, each special event from before now gets its own name, and a separate mapping connects those names to their ID pairs.
        * After this refactor, `SPECIALIZED` means that each different parameter combination gets a dedicated iteration loop. This reduces the number of virtual method calls to check keys that haven't been pressed.
* There is no longer a `robertvec` or `robertmap`. Proper event indexing structs have been introduced as part of object storage.
* `ParsedScope` has been introduced. Tooling in the compiler requests the `ParsedScope *egm_global` rather than `parsed_object*`.
* `parsed_script` is no longer a `parsed_event`. Instead, both of these inherit `ParsedCode`, which points to a `ParsedScope`.
    * `parsed_script` has been renamed to `ParseScript`, as adopting the new style rules is a good way of indicating a recent revisit.
* `parsed_room` no longer pretends to be a `parsed_object`.


azure-pipelines.yml
```diff
@@ -18,6 +18,7 @@ jobs:
   parameters:
     repo: radialgm
     enigmaref: $(Build.SourceBranch)
+    ref: refs/pull/138/merge # Use this for testing a co-dependent rgm pr but comment this before merging
```
fundies, line 21: reminder to remove this before merging. and to update submodule and merge RadialGM/#138 after merging this

azure-pipelines.yml
```diff
@@ -4,6 +4,7 @@ resources:
       type: github
       name: enigma-dev/RadialGM
       endpoint: github.com_EnigmaBot
+      ref: refs/pull/138/merge # Use this for testing a co-dependent rgm pr but comment this before merging
```
fundies, line 7: reminder to remove this before merging. and to update submodule and merge RadialGM/#138 after merging this

fundies:
@JoshDreamland I've fixed the CI for you again. However, I noticed when running gm2egm you aren't writing the EDL to the .edl files and instead putting them in properties.yaml. I believe we fixed this before in one of our 50 branches / PRs but I could be mistaken. We should probably add a gm2egm diffing test as well as a basic EGM read test like we have for the other gmk formats. It should be as simple as converting https://github.com/enigma-dev/enigma-dev/tree/master/CommandLine/testing/SimpleTests/basicGMX.gmx and adding it as an egm to the same folder.

shared/protos/EventDescriptor.proto
```diff
@@ -37,6 +37,8 @@ message EventDescriptor {
     // `constant` should be defined with the engine code to be executed
     // (the former for once-per-frame logic, the latter for per-instance logic).
     INLINE = 5;
+    // ?????
```
fundies, line 40: @JoshDreamland you seem to have missed this type?

fundies:
@JoshDreamland I've fixed egm writing by shuffling some code. Please check my latest commit when you got a chance.

PR #2090, opened 2020-07-26
SDL Joystick Revisions

time-killer-games:
WIP

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
@@ -214,7 +214,6 @@ void SDL_Event_Handler::windowResized(const SDL_Event *event) {
 
 void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) {
   addGamepad(event->cdevice.which);
-  joysticks.resize(SDL_NumJoysticks(), nullptr);
   joysticks_open();
 }
 
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
@@ -1,6 +1,7 @@
 #include "Event.h"
 #include "Window.h"
 #include "Gamepad.h"
+#include "Joystick.h"
 
 #include "Platforms/General/PFmain.h"    //game_end
 #include "Platforms/General/PFwindow.h"  // mouse / keyboard
@@ -211,9 +212,25 @@ void SDL_Event_Handler::windowResized(const SDL_Event *event) {
   enigma_user::draw_clear(enigma_user::window_get_color());
 }
 
-void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) { addGamepad(event->cdevice.which); }
+void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) {
+  addGamepad(event->cdevice.which);
+  int joystick_count = SDL_NumJoysticks();
+  if (joystick_count != joysticks.size()) {
+    joysticks.resize(joystick_count, nullptr);
+  }
+  joysticks_open();
+}
 
-void SDL_Event_Handler::joyDeviceRemoved(const SDL_Event *event) { removeGamepad(event->cdevice.which); }
+void SDL_Event_Handler::joyDeviceRemoved(const SDL_Event *event) {
+  removeGamepad(event->cdevice.which);
+  int joystick_count = SDL_NumJoysticks();
+  if (joystick_count < 0) {
+    joysticks.resize(0);
+  } else {
+    joysticks.resize(joystick_count);
+  }
+  joysticks_close();
+}
 
 void SDL_Event_Handler::controllerButtonDown(const SDL_Event *event) {
   setGamepadButton(event->cdevice.which, event->cbutton.button, true);
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
@@ -220,12 +220,6 @@ void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) {
 
 void SDL_Event_Handler::joyDeviceRemoved(const SDL_Event *event) {
   removeGamepad(event->cdevice.which);
-  int joystick_count = SDL_NumJoysticks();
-  if (joystick_count < 0) {
-    joysticks.resize(0);
-  } else {
-    joysticks.resize(joystick_count);
-  }
   joysticks_close();
 }
 
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
@@ -214,12 +214,12 @@ void SDL_Event_Handler::windowResized(const SDL_Event *event) {
 
 void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) {
   addGamepad(event->cdevice.which);
-  joysticks_open();
+  joysticks_open(event->cdevice.which);
 }
 
 void SDL_Event_Handler::joyDeviceRemoved(const SDL_Event *event) {
   removeGamepad(event->cdevice.which);
-  joysticks_close();
+  joysticks_close(event->cdevice.which);
 }
 
 void SDL_Event_Handler::controllerButtonDown(const SDL_Event *event) {
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
@@ -214,10 +214,7 @@ void SDL_Event_Handler::windowResized(const SDL_Event *event) {
 
 void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) {
   addGamepad(event->cdevice.which);
-  int joystick_count = SDL_NumJoysticks();
-  if (joystick_count != joysticks.size()) {
-    joysticks.resize(joystick_count, nullptr);
-  }
+  joysticks.resize(SDL_NumJoysticks(), nullptr);
   joysticks_open();
 }
 
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
@@ -16,6 +16,7 @@
 **/
 
 #include <string>
+#include <vector>
 
 #include "SDL2/SDL.h"
 #include "Joystick.h"
@@ -28,11 +29,11 @@
 #include "Universal_System/debugscope.h"
 #endif
 
-using enigma::joysticks;
-
 namespace {
 
-static inline SDL_Joystick *joystick_get_handle(int id) {
+std::vector<SDL_Joystick *> joysticks(0);
+
+SDL_Joystick *joystick_get_handle(int id) {
   // joystick id starts at 1 in GameMaker
   #ifdef DEBUG_MODE
   if (id < 1 || id > joysticks.size())) { 
--- a/ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
@@ -109,10 +109,12 @@ void joystick_update() {
   int joystick_count = SDL_NumJoysticks();
   if (joystick_count <= 0) {
     joysticks_close();
+    JOYSTICK_MAX = 0;
+    joysticks.resize(0);
   } else {
     if (joystick_count != JOYSTICK_MAX) {
-      joysticks.resize(JOYSTICK_MAX, nullptr);
       JOYSTICK_MAX = joystick_count;
+      joysticks.resize(JOYSTICK_MAX, nullptr);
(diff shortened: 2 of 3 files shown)


ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -22,11 +22,27 @@
...
 namespace {
 
-SDL_Joystick *joystick1 = NULL;
-SDL_Joystick *joystick2 = NULL;
+SDL_Joystick *joysticks[2] = { NULL, NULL };
 
-static inline SDL_Joystick* joystick_get_handle(int id) {
-  return (id - 1 < 0.5) ? joystick1 : joystick2;
+static inline SDL_Joystick *joystick_get_handle(int id) {
+  return (id - 1 < 0.5) ? joysticks[0] : joysticks[1];
```
fundies, line 28: what are you trying to do here?

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -22,11 +22,27 @@
 
 namespace {
 
-SDL_Joystick *joystick1 = NULL;
-SDL_Joystick *joystick2 = NULL;
+SDL_Joystick *joysticks[2] = { NULL, NULL };
```
fundies, line 25: I know gm only supports 2 joysticks but we can support more by using std::vector<SDL_Joystick*> = {nullptr, nullptr} instead of hard coded array

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -22,11 +22,27 @@
...
 
-static inline SDL_Joystick* joystick_get_handle(int id) {
-  return (id - 1 < 0.5) ? joystick1 : joystick2;
+static inline SDL_Joystick *joystick_get_handle(int id) {
+  return (id - 1 < 0.5) ? joysticks[0] : joysticks[1];
+}
+
+static inline void joysticks_open() {
+  for (size_t i = 0; i < 2; i++) {
```
fundies, line 32: if you replace with vector as I suggested youll want i < joysticks.size() here

time-killer-games:
@fundies alright, i think I did everything you asked correctly, let me know if I didn't or if there's anything else you want done before I ping robert or josh to review and/or merge, I know you don't like macro definitions but that's usually how limits are handled such as PATH_MAX, etc. so I wanted to go with how that sort of thing is usually done. I can change it if you want.

fundies:
> @fundies alright, i think I did everything you asked correctly, let me know if I didn't or if there's anything else you want done before I ping robert or josh to review and/or merge, I know you don't like macro definitions but that's usually how limits are handled such as PATH_MAX, etc. so I wanted to go with how that sort of thing is usually done. I can change it if you want.
> 
> ```
> #define JOYSTICK_MAX 16
> ```
> 
> This way, to change the max number of joysticks only one line needs to be changed. I'm not sure how to make it "not" hard-coded, because this is still technically hard coded but with a bigger number allowable, (i.e. number of gamepads GMS uses).

I don't think you should hard code any value. You should be adding / removing joysticks from the vector when they are plugged / unplugged. I do this with gamepads here: https://github.com/enigma-dev/enigma-dev/blob/master/ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp#L214 

OR: You can also get a raw joystick from a gamepad using sdl's api iirc. This would eliminate need for the extra vector.

time-killer-games:
@fundies i think i found a way to do it that keeps all the changes to one file instead of editing the event handler source. not sure if you find that acceptable, just let me know whenever.

fundies:
> @fundies i think i found a way to do it that keeps all the changes to one file instead of editing the event handler source. not sure if you find that acceptable, just let me know whenever.

I would prefer you do it where I suggested. The code for adding gamepads/joysticks should be in the same place.

time-killer-games:
@fundies alright i think i did it, (again), the more critiques the better, This is nearly the second other third time I ever touched the joystick code, it would be nice to have done right sooner than later. I want to make a habit of doing things correctly the first time going forward and use pull requests as an opportunity to improve my contributions instead of feeling nagged by code review which is counterproductive.

ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
```diff
@@ -211,9 +212,22 @@ void SDL_Event_Handler::windowResized(const SDL_Event *event) {
...
+}
 
-void SDL_Event_Handler::joyDeviceRemoved(const SDL_Event *event) { removeGamepad(event->cdevice.which); }
+void SDL_Event_Handler::joyDeviceRemoved(const SDL_Event *event) {
+  removeGamepad(event->cdevice.which);
+  int joystick_count = SDL_NumJoysticks();
+  if (joystick_count < 0) {
+    joysticks.resize(0);
+  } else {
+    joysticks.resize(joystick_count);
```
fundies, line 227: I don't think you want to be shrinking the vector at all. This will cause issue say if you have 3 joysticks and you remove the 2nd one. You then would have joystick {js1, nullptr, js2} and you would shrink the vector to {js1, nullptr} is not what you want. On add you should iterate the vector to check for the first nullptr and then resize the vector if you need to add additional joysticks.

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -20,13 +20,12 @@
...
+using enigma::joysticks;
 
-SDL_Joystick *joystick1 = NULL;
-SDL_Joystick *joystick2 = NULL;
+namespace {
 
-static inline SDL_Joystick* joystick_get_handle(int id) {
-  return (id - 1 < 0.5) ? joystick1 : joystick2;
+static inline SDL_Joystick *joystick_get_handle(int id) {
+  return (id < joysticks.size()) ? ((id < 1) ? joysticks[0] : joysticks[id - 1]) : joysticks[joysticks.size() - 1];
```
fundies, line 28: I'm not a big fan of this check I would prefer you instead do
```cpp
#ifdef DEBUG_MODE
if (id < 0 || id >= joysticks.size())) { DEBUG_MESSAGE("invalid js id"); return nullptr}
#endif
return joysticks[id];
```

You could also use an AssetArray which does these kind of checks for you. Either would be fine with me.

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -38,22 +37,19 @@ bool joystick_exists(int id) {
 }
 
 std::string joystick_name(int id) {
-  const char *name;
-  name = SDL_JoystickName(joystick_get_handle(id));
+  const char *name = SDL_JoystickName(joystick_get_handle(id));
   return (name ? name : "");
```
fundies, line 41: nit: ```name == nullptr``` for clarity

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -38,22 +37,19 @@ bool joystick_exists(int id) {
...
+  int axes = SDL_JoystickNumAxes(joystick);
   return (axes > 0 ? axes : 0);
 }
 
 int joystick_buttons(int id) {
-  int buttons;
   SDL_Joystick *joystick = joystick_get_handle(id);
-  buttons = SDL_JoystickNumButtons(joystick);
+  int buttons = SDL_JoystickNumButtons(joystick);
   return (buttons > 0 ? buttons : 0);
```
fundies, line 53: If a negative value is returned it means something has gone wrong. I think we should report this in debug mode.
```cpp
#ifdef DEBUG_MODE
if (buttons < 0) DEBUG_MESSAGE("Failed to query jostick id  + SDL_GetError());
#endif
```

ENIGMAsystem/SHELL/Platforms/SDL/Event.cpp
```diff
@@ -211,9 +212,22 @@ void SDL_Event_Handler::windowResized(const SDL_Event *event) {
   enigma_user::draw_clear(enigma_user::window_get_color());
 }
 
-void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) { addGamepad(event->cdevice.which); }
+void SDL_Event_Handler::joyDeviceAdded(const SDL_Event *event) {
```
fundies, line 215: As a minimal addition to your pr it might be nice to print something  like "Joystick 3 ("DUAL SHOCK 3") connected/disconnected" in debug mode for these events

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
time-killer-games: To remain GameMaker compatible we will need the joystick id to start at 1, not zero, other than that, I 100% agree with you. I don't like it starting at 1 though, I wish GM didn't do that.

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
time-killer-games: Do we want to print any DEBUG_MESSAGE's with the contents of SDL_GetError()? This would a nice application to use that, in case the joystick name can't be retrieved for whatever reason, it would print more specifically what the exact problem was that we might not know without peeking at SDL's source and making our own errors, which would be a waste of time imo, since errors are already provided by the api.

https://wiki.libsdl.org/SDL_GetError

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
fundies: You can change my code to reflect gm's off by 1 bug like so

```cpp
#ifdef DEBUG_MODE
if (id < 1 || id > joysticks.size())) { DEBUG_MESSAGE("invalid js id"); return nullptr}
#endif
return joysticks[id-1];```
but please comment why you do that in the src.

However, I'm not a big fan of emulating GM's bugs in our code base just to maintain compatibility. I think a good compromise might be adding a function or variable to set first js id offset id based on some sort gm compatibility settings / extensions we add.

ie something like

```cpp 
namespace enigma {
const size_t js_offset = 1;
}
````
```cpp
#include whater_js.h

inline SDL_Joystick *joystick_get_handle(int id) {
  #ifdef DEBUG_MODE
  if (id < js_offset || id >= joysticks.size() + js_offset)) { DEBUG_MESSAGE("invalid js id"); return nullptr}
  #endif
  return joysticks[id-js_offset];```
}
```
However, I don't really expect you to handle this in this pr so just leaving a comment for now explaining why you did what you did will suffice for now I guess.

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
fundies: if you fail to get the name you can print a debug message here too if you wish. SDL_GetErrror() just returns const char* so you should be able to use it directly in DEBUG_MESSAGE.  However, you will need a std::string if you want to concat it.
ie ```DEBUG_MESSAGE(std::string("Error: enigma took a shit ")  + SDL_GetError());```

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
```diff
@@ -77,48 +80,37 @@ double joystick_axis(int id, int numb) {
...
-  if (joystick_count <= 0) {
-    if (joystick1 != NULL) {
-      SDL_JoystickClose(joystick1);
-      joystick1 = NULL;
-    }
-
-    if (joystick2 != NULL) {
-      SDL_JoystickClose(joystick2);
-      joystick2 = NULL;
+void joysticks_open() {
```
fundies, line 96: This will work for first run but as soon as you start plugging / unplugging things will get messy. You should be passing the joystick id from the event and using that to call open/close. 

https://github.com/enigma-dev/enigma-dev/blob/master/ENIGMAsystem/SHELL/Platforms/SDL/Gamepad.cpp#L38

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
time-killer-games: You mean opening and closing based on this? https://github.com/enigma-dev/enigma-dev/blob/5881bfb3a6eb53b67d30f2b7ae6fd72211a69ca1/ENIGMAsystem/SHELL/Platforms/SDL/Gamepad.cpp#L55

ENIGMAsystem/SHELL/Platforms/SDL/Joystick.cpp
time-killer-games: You mean open and close based on this? https://github.com/enigma-dev/enigma-dev/blob/5881bfb3a6eb53b67d30f2b7ae6fd72211a69ca1/ENIGMAsystem/SHELL/Platforms/SDL/Gamepad.cpp#L55

fundies: No. I mean look at my logic for adding / removing from gamepads vector. your code should be a mirror of it

time-killer-games: @fundies I'm not going to test it just yet because there are other things I have to do atm, but feel free to point out anything I probably did wrong until I'm available again to address it.

PR #2104, opened 2020-08-09
Finish EgmEvents

JoshDreamland:
Corrects No More Health and replaces event prefixes/suffixes with a proper event dispatcher mechanism. This is all I'm aware of in way of regressions from fundies merging the other events branch.


fundies:
I'm now seeing ```Yaml2Proto: Can't find field `Suffix` in message `EventDescriptor`...```

[1945.gm6.zip](https://github.com/enigma-dev/enigma-dev/files/5048234/1945.gm6.zip) Seems to have an issue in DND conversion but even after fixing it manually in the egm the game does not seem to run correctly.

PR #2105, opened 2020-08-09
fix default font

fundies:


ENIGMAsystem/SHELL/Graphics_Systems/General/GSfont.cpp
```diff
@@ -133,7 +107,7 @@ namespace enigma_user {
 double string_char_width(variant vstr)
 {
   string str = toString(vstr);
-  get_font(fnt,currentfont,0);
+  SpriteFont& fnt = sprite_fonts[currentfont];
```
RobertBColton, line 110: This would default construct a font for -2 and everything else and not just the default font. This completely destroys the ability to see if a font actually exists. I'm not sure the macros were correct either though, I believe my original intention was that AssetArray::get was supposed to perform bounds checking like STL does, and all asset functions like this would be required to get the resource that way.

ENIGMAsystem/SHELL/Graphics_Systems/General/GSfont.cpp
```diff
@@ -1287,28 +1261,34 @@ void draw_text_ext_color(gs_scalar x, gs_scalar y,variant vstr,gs_scalar sep, gs
...
 }
 
 void draw_set_font(int fnt) {
-  enigma::currentfont = fnt;
+  if (fnt == -1 || sprite_fonts.exists(fnt))
+    enigma::currentfont = fnt;
+  #ifdef DEBUG_MODE
+  else
+    DEBUG_MESSAGE("Requested font asset " + std::to_string(fnt) + " does not exist.", MESSAGE_TYPE::M_USER_ERROR);
+  #endif
```
RobertBColton, line 1290: Here is an alternative that works if something else happens besides a delete.
```cpp
struct CurrentFont {
  CurrentFont(int id): _id(id) {}
  operator=(int other) { _id = other; }
  operator int() { 
    if (!sprite_fonts.exists(_id)) return -1;
    return _id;
  }
} currentfont;
```

ENIGMAsystem/SHELL/Universal_System/Resources/fontstruct.cpp
```diff
@@ -224,6 +225,12 @@ string font_get_fontname(int fnt)
 
 void font_delete(int fnt)
 {
+    if (enigma::currentfont == fnt) {
+       enigma::currentfont = -1;
+       #ifdef DEBUG_MODE
+       DEBUG_MESSAGE("Deleted currently in use font " + std::to_string(fnt) , MESSAGE_TYPE::M_USER_ERROR);
+       #endif
+    }
```
RobertBColton, line 233: Here is an alternative that works if something else happens besides a delete.
```cpp
struct CurrentFont {
  CurrentFont(int id): _id(id) {}
  operator=(int other) { _id = other; }
  operator int() { 
    if (!sprite_fonts.exists(_id)) return -1;
    return _id;
  }
} currentfont;
```

ENIGMAsystem/SHELL/Graphics_Systems/General/GSfont.cpp
RobertBColton, line 1290: GM8.1 and GMSv1.4 will not show this for font -1, -2, or any other non-existing font. However, I am ok with this change because the manual only says that -1 is acceptable for the default font.

ENIGMAsystem/SHELL/Graphics_Systems/General/GSfont.cpp
```diff
@@ -1287,28 +1261,34 @@ void draw_text_ext_color(gs_scalar x, gs_scalar y,variant vstr,gs_scalar sep, gs
   }
 }
 
-unsigned font_get_texture(int fnt) {
-  get_font_null(f,fnt,-1);
-  return f.texture;
+unsigned font_get_texture(int id) {
+  const SpriteFont& fnt = sprite_fonts.get(id);
+  return fnt.texture;
 }
```
RobertBColton, line 1267: The following shows `<undefined>` and `0` for GM8.1 and GMSv1.4 in the creation event. So I am totally fine with this, everything that take a font, doesn't accept anything but user fonts. I'm totally fine with that.

GMSv1.4 shows this debug message for -2, -1, and 0 fonts. So we're doing the correct thing here now.
Font Doesn't Exist


```cpp
show_message(font_get_name(-2));
show_message(font_get_name(-1));
show_message(font_get_name(0));

show_message(font_get_texture(-2));
show_message(font_get_texture(-1));
show_message(font_get_texture(0));
```

ENIGMAsystem/SHELL/Universal_System/Resources/fontstruct.cpp
RobertBColton: Ok, so in every other respect this pull request is just like GMSv1.4 except for font deletion and possibly the current font. I don't think it's a good idea that we allow deletion of the default font... but I suppose we can allow it. However, GMSv1.4 does not allow it, just like all of the other functions which accept a font id.

GMSv1.4 Delete Default Font Error

RobertBColton: I made a little test for this specifically. Just creating a new font and setting its size to 56 before running. It does not produce a debug message in GMSv1.4 like it does this pull request. However, it does show that GM properly recognizes the current font being deleted and updates accordingly.
```cpp
draw_set_font(font0);
show_message(string(string_width("hellooooooo"))); // 412
font_delete(font0);
show_message(string(string_width("hellooooooo"))); // 102
```

PR #2138, opened 2020-09-11
Optionally Hide Debug Output (Finished Games Only)

time-killer-games:
It would be desirable to hide all debug output the engine does internally to give the user complete control of what gets printed and what doesn't. 

Intended for complete games. Should never be used otherwise so bugs can be debugged.

HIDDEN_DEBUG_OUTPUT needs to be defined by a checkbox in LateralGM/RadialGM and a flag in emake for these changes to even be usable.


fundies:
I would instead change the implementations in the widget systems. Users may want to turn these messages on and off at certain points too. We could also allow this to be more granular. I added message types for user_error, user_warning and etc but you could even add more so users could disable/hide all sprite lookup failures or all errors from a certain object type or whatever.

```cpp
std::map<message_type, bool> printErrs;
set_debug_output_enabled(bool enabled, message_type) {
  printErrs[message_type] = enabled;
}

void show_debug_message(string message, message_type) {
  if (!printErrs[message_type]) return;
  ...
} 
```

time-killer-games:
@fundies that's a good idea. @JoshDreamland ignore my message and please don't merge this, i didn't see fundies review until just now

ENIGMAsystem/SHELL/Widget_Systems/widgets_mandatory.h
```diff
@@ -23,7 +23,21 @@
 
 #include <string>
 
-#define DEBUG_MESSAGE(msg, severity) ::enigma_user::show_debug_message((std::string) (msg) + " | " __FILE__ ":" + std::to_string(__LINE__), (severity))
+#ifdef ABORT_ALL_ERRORS
+  #ifdef HIDE_DEBUG_MESSAGES
+    #define DEBUG_MESSAGE(msg, severity) if (severity >= 2) { exit(1); } \
```
JoshDreamland, line 28: The debug message won't be printed for severe errors. It'll just silently exit. Also, you should use a constant instead of 2.

ENIGMAsystem/SHELL/Widget_Systems/widgets_mandatory.h
```diff
@@ -23,7 +23,21 @@
 
 #include <string>
 
-#define DEBUG_MESSAGE(msg, severity) ::enigma_user::show_debug_message((std::string) (msg) + " | " __FILE__ ":" + std::to_string(__LINE__), (severity))
+#ifdef ABORT_ALL_ERRORS
```
JoshDreamland, line 26: Instead of duplicating the macro for all permutations of these two variables, I'd ensure ABORT_ON_ALL_ERRORS was defined to false by default and then include it in the if().

JoshDreamland (changes requested):

time-killer-games:
@fundies i'd still like to have a checkbox for the ide and flag for emake so this couldd be overriden for an entire project simply by the click of a mouse and no code editing. But there you have it, we have your functions in here as well

ENIGMAsystem/SHELL/Widget_Systems/widgets_mandatory.h
time-killer-games: @JoshDreamland is what i have no ok? for both things you addressed

fundies:
> @fundies i'd still like to have a checkbox for the ide and flag for emake so this could be overridden for an entire project simply by the click of a mouse and no code editing. But there you have it, we have your functions in here as well
> 
> I'd be happy to add these things myself but i only intend to do it with rgm and emake and won't until i need it.

Adding check boxes to ides and etc is a pain. Even though I wouldn't consider this an extension you could create one that enables this for you using a static intializer. I do this in the freetype extension. See FontManager class for an example.

time-killer-games:
My initial thought was to make it an extension but I wasnt sure if you guys would like that idea and not merge it I added it.

fundies:
> My initial thought was to make it an extension but I wasnt sure if you guys would like that idea and not merge it I added it.

We'll the code I mentioned should only be a few lines. The res should be in universal. I do not think it should permanently live there but I think it is really your only option aside from hacking up the compiler. I think we need a better mechanism for conditional contemplation of various settings but for now this all we got.

PR #2139, opened 2020-09-12
Let Win32 EXE run from paths w/ foreign characters

time-killer-games:
"Resource load fail: exe unopenable" error will be thrown if your exe has a file name or absolute path it is inside of containing non-English characters such as Cyrillic.

This pr will fix that when it's done.

### DIFF
--- a/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
+++ b/ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSmain.cpp
@@ -61,7 +61,7 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam);
 
 void (*touch_extension_register)(HWND hWnd);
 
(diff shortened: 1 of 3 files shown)


time-killer-games:
The code in its current start throws an error at this line:
https://github.com/enigma-dev/enigma-dev/blob/05bab38d558245506422405ff3e9e09d1aaa4c86/ENIGMAsystem/SHELL/Universal_System/Resources/loading.cpp#L108

@JoshDreamland @RobertBColton @fundies any idea why?

fundies:
Just fyi. I don't get ping or notifications from you while. I'm blocked. Also do we need this widen/shorten crap still. Can't std::filesystem handle this yet?

time-killer-games:
we do unfortunately, in the amount of time we could spend adding the c++17 implementation it will already be should be removed in favor of c++20. Where as with the Win32 shorten/widen will never go stale based on microsoft's track record and the fact they dont openly say they plan to depreciate it any time soon (if ever).

PR #2145, opened 2020-10-29
Switch to opengl 3 as default

fundies:
Switched to OpenGL 3 as default because:

a) OpenGL 1.1 is prehistoric and we should phase it out for the more compatible GLES2
b) Our OpenGL 1.1 system is calling OpenGL 3 functions in many places that would require further refactoring of an already dead system.


JoshDreamland (approved):

JoshDreamland (approved):
Unfortunate ending for GL1, but possibly for the best. Do we know if GL3 works for hugar?

Hughtech:
switching to opengles2 is a proper move.  However i would still keep opengl1 as a fallback.....perhaps for debug/legacy purposes. As for me, I recently acquired a graphics card so it should be able to run opengles2........i would have to test it out to make sure.

PR #2167, opened 2020-11-25
New Parser [NOT READY]

JoshDreamland:
Step one: Remove dependencies on JDI from the Lexer.

This gives ENIGMA a custom representation of JDI macros. This functionality is untested. Actually, the only test for the lexer right now is misnamed, but I need to shelve this and fix something in RGM.

This commit integrates the new lexer into the old parser by re-stringifying the lex after syntax checking. It replaces the syntax checker with a call to AST::ParseAST. **However, this method is *NOT YET IMPLEMENTED,*** meaning ***no syntax checking happens** as of this commit.*

In all likelihood there are remaining problems tokenizing strings, etc. Also, the `INCREMENT` and `DECREMENT` tokens cannot be programmatically removed from the token trie at this time.

Theoretically, though, valid games should still build as of this commit. I expect the CI to suggest otherwise.

BEFORE WE COMMIT THIS:  The missing methods on AST must be authored, building the actual tree. This change puts the compiler where I *thought* it was when I started this task.


CompilerSource/parsing/macros.cpp
```diff
@@ -0,0 +1,316 @@
...
+      if (!i) {
+        herr->Error(tokens[i]) << "Concatenation marker (`##`) "
+            "cannot appear at either end of a replacement list.";
+        continue;
+      }
+      // Don't push spans CONCAT would always discard.
+      if (meaningful_span(tokens, e, i))
+        res.push_back(FuncComponent::TokenSpan{e, i});
+      res.push_back(FuncComponent::Paste{});
+      while (++i < tokens.size() && tokens[i].PreprocessesAway());
```
fundies, line 105: i thought this type of syntax caused a warning in gcc? plus isn't it less confusing for readers to just put the continue?

CompilerSource/parsing/macros.cpp
```diff
@@ -0,0 +1,316 @@
...
+  Token res = l.ReadRawToken();
+  Token end = l.ReadRawToken();
+  if (end.type != TT_ENDOFCODE) {
+    errc.Error() << "Concatenation of `" << left.content
+                 << "` and `" << right.content
+                 << "` does not produce a valid token.";
+  }
+  return res;
+}
+/*
```
fundies, line 190: Maybe you're fixing this soon but maybe you'll forget. IDK either way I think its a good practice to leave a comment for why its here in case you forget and we end up with a new copy later

CompilerSource/parsing/macros.cpp
```diff
@@ -0,0 +1,316 @@
...
+    }
+  }
+  return res;
+}
+*/
+static void AppendOrPaste(TokenVector &dest,
+                          TokenVector::const_iterator begin,
+                          TokenVector::const_iterator end,
+                          bool paste, const ErrorContext &errc) {
+  if (begin == end) return;
```
fundies, line 217: is >= possible? maybe we should warn when this happens?

CompilerSource/parsing/macros.cpp
```diff
@@ -0,0 +1,316 @@
...
+  dest.insert(dest.end(), begin, end);
+}
+
+TokenVector Macro::SubstituteAndUnroll(
+    const vector<TokenVector> &args, const vector<TokenVector> &args_evald,
+    ErrorContext errc) const {
+  TokenVector res;
+  bool paste_next = false;
+  if (args.size() < parameters->size()) {
+    if (!is_variadic || args.size() + 1 < parameters->size()) {
```
fundies, line 235: im confused at what the difference is between this snipped and the one directly below it

CompilerSource/parsing/macros.cpp
```diff
@@ -0,0 +1,316 @@
...
+            part.tag == FuncComponent::EXPANDED_ARGUMENT
+                ? args_evald[ind] : args[ind];
+        AppendOrPaste(res, vec.begin(), vec.end(), paste_next, errc);
+        paste_next = false;
+        break;
+      }
+      case FuncComponent::PASTE:
+        paste_next = true;
+        break;
+      case FuncComponent::VA_OPT: {
```
fundies, line 297: It might be helpful to have some comments on what each of the switch cases are used for exactly

CompilerSource/parsing/macros.h
```diff
@@ -0,0 +1,142 @@
...
+
+#include <map>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace enigma {
+namespace parsing {
+
+typedef std::vector<Token> TokenVector;
```
fundies, line 32: maybe we should start to prefer using "using" ?

CompilerSource/parsing/macros.h
```diff
@@ -0,0 +1,142 @@
...
+      RAW_ARGUMENT,
+      EXPANDED_ARGUMENT,
+      PASTE,
+      STRINGIFY,
+      VA_OPT
+    };
+    struct TokenSpan {
+      size_t begin, end;
+    };
+    struct Argument {
```
fundies, line 61: why?

CompilerSource/parsing/tokens.h
```diff
@@ -0,0 +1,150 @@
...
+  TT_NOTEQUAL,        // The != or <> operator.
+  TT_BANG,            // The ! operator.
+  TT_TILDE,           // The ~ operator.
+  TT_INCREMENT,       // The ++ operator.
+  TT_DECREMENT,       // The -- operator.
+  TT_LESS,            // The < operator (or opening angle bracket).
+  TT_GREATER,         // The > operator (or closing angle bracket).
+  TT_LESSEQUAL,       // The <= operator.
+  TT_GREATEREQUAL,    // The >= operator.
+  TT_LSH,             // The << operator.
```
fundies, line 66: glad you commented these

CompilerSource/parsing/tokens.h
```diff
@@ -0,0 +1,150 @@
...
+  TTM_WHITESPACE,     // Preprocessing token: whitespace
+  TTM_CONCAT,         // Preprocessing token: ##
+  TTM_STRINGIFY,      // Preprocessing token: #
+
+  TT_ERROR,
+  TT_ENDOFCODE
+};
+
+namespace internal {
+namespace useless {
```
fundies, line 113: Ok but why is it useless? perhaps leave a reason for why its here so we can phase it out later

shared/eyaml/eyaml.cpp
```diff
@@ -174,6 +174,7 @@ struct y_level {
   y_level(y_level* a, ey_data *b, int c): prev(a), s(b), i(c) {}
 };
 
+typedef size_t pt;
```
fundies, line 177: ive seen this in a few places.isn't there a common header it can be done in?

fundies (changes requested):
I'll look into the build errors but heres a few nits from me

CompilerSource/parsing/ast.h
```diff
@@ -0,0 +1,67 @@
...
+  // Utility routine: Apply this AST to a specified instance.
+  void ApplyTo(int instance_id);
+
+  // Parses the given code, returning an AST*. The resulting AST* is never null.
+  // If syntax errors were encountered, they are stored within the AST.
+  static AST Parse(std::string code, const ParseContext *ctex);
+
+  // Disallow copy. Our tokens point into our code.
+  AST(const AST &) = delete;
+  AST(AST &&other);
```
fundies, line 55: = delete?

PR #2174, opened 2020-12-15
Add Platform os_* enum to SDL

time-killer-games:


time-killer-games:
@JoshDreamland this one is ready. Thanks!

fundies:
I don't understand what this fixes. Did you have a build error?

time-killer-games:
@fundies os_* constants aren't included in SDL therefore they are not usable for that platform currently. This allows you to do platform-specific checks and platform specific code execution in SDL like we currently can with other platforms. Although I forgot something, it needs to have a -Dos_type=whatever in the makefile which I'm going to add right now.

time-killer-games:
@JoshDreamland ok correction, now this one is actually ready

fundies:
> @fundies os_* constants aren't included in SDL therefore they are not usable for that platform currently. This allows you to do platform-specific checks and platform specific code execution in SDL like we currently can with other platforms. Although I forgot something, it needs to have a -Dos_type=whatever in the makefile which I'm going to add right now.

But adding them to a cpp won't make the accessible in the engine :/

JoshDreamland:
Is that constant declared elsewhere? AFAICT you've deleted the only declaration of it that would be available to users.

time-killer-games:
> Is that constant declared elsewhere? AFAICT you've deleted the only declaration of it that would be available to users.

@JoshDreamland @fundies I was wondering about that. Initially i removed it because it was the source of a compile error but now i realize the error only happened because it was defined as a const and I forgot the const for the code I added so that's a const/non-const type conflict. Fixing that now, I techincally added the const already so I'll just add back the extern.

time-killer-games:
@JoshDreamland mudge kk

ENIGMAsystem/SHELL/Platforms/General/PFsystem.h
```diff
@@ -36,6 +36,7 @@ enum {
   os_winphone = 7,
   os_tizen = 8,
   os_win8native = 9,
+  os_bsd = 10,
```
fundies, line 39: You're not supposed to have comma on last enum. I think some compilers will cry or throw warnings

fundies:
I don't know about your placement of these variables as you're duplicating them between SL and Win32. I feel like these cpps belong somewhere they both can share the declaration like bridges maybe?

ENIGMAsystem/SHELL/Platforms/General/PFsystem.h
time-killer-games: i didn't know that mainly because that's how it already was before i created this pr, but thanks for the info better to correct that now than later.

time-killer-games:
Agreed. Bridges is the way to go. The more platforms we add in the current layout the messier and more that would need to be corrected over time. Better to just move it under bridges now to end the madness. lel

time-killer-games:
That said, will close this and make the new pr as i would just have to undo all changes in this one anyway

PR #2178, opened 2020-12-23
Dialog Update 2020

time-killer-games:
TODO:
- Cocoa
- xlib


time-killer-games:
@JoshDreamland I have it building and running OK in Run/Debug mode. However compile mode does this: https://pastebin.com/mGm6GHVb

No idea how I'm supposed to fix this. Do you?

time-killer-games:
@JoshDreamland I still haven't figured out a way to get rid of those errors, despite fixing a few imperfects i found and resolved in the more recent commits since my last comment. However, I do have the Run Mode executable still working as mentioned before, which i retrieved out of my msys2 tmp folder and have uploaded in this post below.

[DialogModule.zip](https://github.com/enigma-dev/enigma-dev/files/5740503/DialogModule.zip)

You can test that file with WINE if you want since you probably don't have access to Windows easily, however, the input box dialogs don't show up in WINE and are skipped since they require API's that the WINE developers have not implemented yet, probably due the fact it relies on a System32 ActiveX COM object which is not portable. This is not a problem on Windows, obviously. Everything else should work in WINE out of the box.

JoshDreamland:
Versioning the binary as text has all the same problems as versioning it as a binary. Changes to it clog up the commit history. You could split a single method into two and the entire file would be different.

Also, as text, this change is bigger than the entire remainder of the engine...

PR #2179, opened 2020-12-29
unix macro not defined on linux dumb dumbs

time-killer-games:
But linux macro is defined on linux, meaning it won't even make it to that elif check on linux either way


time-killer-games:
@JoshDreamland restart crappveyor

PR #2186, opened 2021-02-14
Update file IO command line test: Renamed function 'directory_delete'

van0014:
Change function name 'directory_delete' to 'directory_destroy' at line 206, 
as discovered by gfundies and TKG


JoshDreamland (approved):
Cool; thanks.

PR #2190, opened 2021-03-11
First Contribution

pushkar7570:
Webchat Freenode channel name is changed from ``#enigma-dev``  to  ``#enigma``  because the link http://webchat.freenode.net/?channels=enigma&uio=d4  leads you to the channel  ``#enigma``  and the channel named ``#enigma-dev``  does not have any participants.


Readme.md
```diff
@@ -6,7 +6,7 @@ If you find our tools useful, then please consider supporting ENIGMA on Patreon!
 
 Official Website: http://enigma-dev.org
 
-ENIGMA, the Extensible Non-Interpreted Game Maker Augmentation, is an open-source, cross-platform game development environment based on the popular software, Game Maker. Its intent is to provide users with a quality game creation tool and a bridge between high- and low-level programming languages. It can be used either through an IDE, such as its sister project, [LateralGM](https://github.com/IsmAvatar/LateralGM), or its up-and-coming custom frontend, [RadialGM](https://github.com/enigma-dev/RadialGM), or also through its command line interface, [emake](CommandLine/emake).
```
JoshDreamland, line 9: This hyphen's correct, actually; it's a style thing. Another example is [this Wikipedia article](https://en.wikipedia.org/wiki/High-_and_low-level).

JoshDreamland (commented):
Good catch! The hyphen was right, though. 😉

Readme.md
pushkar7570: sure i will make this correct in next pull request, actually i am a newbie 😜

JoshDreamland (approved):
You're doing fine. Thanks for the patch!

Issue #2191, opened 2021-03-12
X11 Widgets are broken (X11/Zenity and Everything/Wayland)


time-killer-games:
This is a reminder for me to fix X11 widgets on newer systems thanks to an update to zenity by its maintainers that broke it (they removed the ability to attach the dialog transient for an external window, now it doesnt show any dialogs crying about an unknown option because they dropped it)

I have a workaround in place for this. But there's another problem: Wayland can't attach windows transient for foreign Windows belonging to other processes. This I also have a fix for, it involves forcing the dialogs to use X11 instead of wayland. This will allow them to stay on top of the game window and not force you to alt tab because the game window will not be able to go on top of the dialogs.

The latter problem fortunately is fixable with only one line of code to add/change: https://github.com/time-killer-games/libdlgmod/blob/e5df435792a930051155a7ed84bff8df7878cb64/DlgModule/xlib/dlgmodule.cpp#L111

Now with Zenity's issue I need to remove all references to the --attach flag and wammo it no longer cries unknown option.

PR #2192, opened 2021-03-12
fix x11 widgets (Zenity update from maintainers and Wayland Sessions broke stuff)

time-killer-games:
fixes one out of two problems mentioned in https://github.com/enigma-dev/enigma-dev/issues/2191 in particular the latter one mentioned by simply disabling the wayland display, forcing the x11 fallback on zenity and kdialog, since they will inherit the environment variable from parent process. Fixes kdialog.

For zenity to be fixed, this is just part of the problem. That is a WIP atm.

Edit:

This pr is ready for review once the CI passes.

### DIFF
--- a/CI/solve_engine_deps.sh
+++ b/CI/solve_engine_deps.sh
@@ -33,7 +33,7 @@ if [ "$WIDGETS" == "GTK+" ] || [ "$TEST_HARNESS" == true ]; then
   LINUX_DEPS="$LINUX_DEPS libgtk2.0-dev"
 fi
 if [ "$WIDGETS" == "xlib" ] || [ "$TEST_HARNESS" == true ]; then
-  LINUX_DEPS="$LINUX_DEPS zenity kdialog libprocps-dev"
+  LINUX_DEPS="$LINUX_DEPS zenity kdialog"
 fi
 
 ###### Extensions #######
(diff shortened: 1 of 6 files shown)


time-killer-games (commented):
@JoshDreamland libprocps-dev is just the development headers. On Ubuntu and Debian, the related libraries themselves are already installed, so this doesn't add any new dependency for the end users. Every desktop environment that I can think of has this dependency, and Ubuntu/Debian is one of the odd few which don't have the development headers installed by default, where as all other major distro's have the headers installed alongside the libraries. This also goes for Arch Linux/Gentoo as any DE you could possibly install on it will have this dependency, unless you happen to be ok wirh using raw Xorg's twm with no DE, chances are Arch/Gentoo systems will have this library installed as well.

JoshDreamland (approved):
I'm very pleasantly surprised, right now; this code reads as though it went through a code review since the last time I looked at it. You've even got `static` before methods that don't need to be used outside the translation unit. File descriptors are cleaned up even when `fork()` fails, and the numerous changes are all very clearly correct (simply `shellscript_evaluate` → `create_shell_dialog`). If you'd shown me those two commits, I'd have said they were written by different people. Good work!

time-killer-games:
@JoshDreamland thank you josh! The function with the pipes and file desciptors i got from a stackoverflow answer, but since the original time i have copied and pasted that function i have looked up the docs to verify it was done correctly because i was skeptical at first.

PR #2198, opened 2021-03-15
umm, i can explain

time-killer-games:
tl;dr. In the process of me trying to fix dialogs, I accidentally broke them in a different way, while not actually fixing anything. 

https://github.com/enigma-dev/enigma-dev/commit/a019ab6c06cd922d264abef7f0f3a154d66c867d

When ci passes please merge. Posix threads are a more technically correct approach than using a fork anyway.

On Unix likes, regardless of the default shell executable that's actually being used, getting the first argument in the command line will return "/bin/sh" and this has been confirmed from testing. The problem was I wasnt getting a pid match with the launched process and kdialog/zenity because it was giving me the process id of the default shell instead; this code filters out the default shell to ensure a pid match so we can verify the topmost window is owned by the correct process.

### DIFF
--- a/ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
+++ b/ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
@@ -196,6 +196,27 @@ static bool ProcIdExists(pid_t procId) {
   return (kill(procId, 0) == 0);
 }
 
+static void ParentProcIdFromProcId(pid_t procId, pid_t *parentProcId) {
+  #if CURRENT_PLATFORM_ID == OS_MACOSX
+  proc_bsdinfo proc_info;
+  if (proc_pidinfo(procId, PROC_PIDTBSDINFO, 0, &proc_info, sizeof(proc_info)) > 0) {
+    *parentProcId = proc_info.pbi_ppid;
+  }
+  #elif CURRENT_PLATFORM_ID == OS_LINUX
+  PROCTAB *proc = openproc(PROC_FILLSTATUS | PROC_PID, &procId);
+  if (proc_t *proc_info = readproc(proc, nullptr)) { 
+    *parentProcId = proc_info->ppid;
+    freeproc(proc_info);
+  }
+  closeproc(proc);
+  #elif CURRENT_PLATFORM_ID == OS_FREEBSD
+  if (kinfo_proc *proc_info = kinfo_getproc(procId)) {
+    *parentProcId = proc_info->ki_ppid;
+    free(proc_info);
+  }
+  #endif
+}
+
 static void ProcIdFromParentProcId(pid_t parentProcId, pid_t **procId, int *size) {
   std::vector<pid_t> vec; int i = 0;
   #if CURRENT_PLATFORM_ID == OS_MACOSX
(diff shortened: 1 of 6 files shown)


time-killer-games:
@JoshDreamland aight stinky poo brain, mudge pls

time-killer-games:
@JoshDreamland hey boi

time-killer-games:
@fundies @JoshDreamland umm don't cancel my ci kthx

fundies:
> @fundies @JoshDreamland umm don't cancel my ci kthx

I was doing things and didn't want to wait. You can always close/reopen to trigger it again

time-killer-games:
It's ok, I was just asking if it could wait this time if the same were to occur this time around.

ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
```diff
@@ -146,14 +161,86 @@ static pid_t process_execute(const char *command, int *infp, int *outfp) {
...
+  int cntp; if (kinfo_proc *proc_info = kinfo_getallproc(&cntp)) {
+    for (int j = 0; j < cntp; j++) {
+      if (proc_info[j].ki_ppid == parentProcId) {
+        vec.push_back(proc_info[j].ki_pid); i++;
+      }
+    }
+    free(proc_info);
+  }
+  #endif
+  *procId = (pid_t *)malloc(sizeof(pid_t) * vec.size());
```
JoshDreamland, line 206: Why are you copying the vector to a new buffer rather than just returning the vector?

JoshDreamland (commented):
We can't keep adding dependency after dependency to accommodate something like displaying text boxes and prompting for strings. We'll go ahead and add this one, but if we get a bunch of user complaints about the new dep, I'm going to revert the default widget system back to None.

ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
```diff
@@ -146,14 +161,86 @@ static pid_t process_execute(const char *command, int *infp, int *outfp) {
...
-  while (pid_from_wid(display, wid) != pid) {
-    wid = wid_from_top(display);
+  Display *display = XOpenDisplay(nullptr);
+  Window wid = WidFromTop(display);
+  pid_t child = PidFromPpidRecursive((pid_t)(std::intptr_t)pid);
+  while (true) {
+    wid = WidFromTop(display);
+    if (PidFromWid(display, wid) == child) {
+      break;
+    }
```
JoshDreamland, line 234: This looks like a busy wait; I recommend adding a sleep or yield to avoid tying up the CPU while the window launches.

time-killer-games:
> We can't keep adding dependency after dependency to accommodate something like displaying text boxes and prompting for strings. We'll go ahead and add this one, but if we get a bunch of user complaints about the new dep, I'm going to revert the default widget system back to None.

I don't see the pr changing the default widget system to xlib. If my previous or did that, it was a mistake on my end. I meant to keep it sent to None, and I wouldn't have insisted on you merging it in that state if I knew it did that, because I thought we already had it accomplished setting it to None was the proper thing to do.

ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
time-killer-games: Is five milliseconds good?

ENIGMAsystem/SHELL/Widget_Systems/xlib/dialogs.cpp
time-killer-games: Sorry, this is based on code where i was already doing that and I forgot to make it C++ abi and not C.

JoshDreamland (approved):
Probably just got committed by accident at some point along the way. Without forcing this onto users as a default, I have no problems including it as a dep. Users actually expecting to launch dialogs this way will already need to install Zenity/KDialog, so this extra library isn't a big deal.

PR #2201, opened 2021-03-17
i did it!

time-killer-games:
I'm not sure what i did

### DIFF
--- a/CompilerSource/general/bettersystem.cpp
+++ b/CompilerSource/general/bettersystem.cpp
@@ -252,7 +252,11 @@ void myReplace(std::string& str, const std::string& oldStr, const std::string& n
       return e_exec(cmd, eCenviron);
     }
 #else
+    #include <vector>
+    #include <algorithm>
+    #include <cstdlib>
     #include <csignal>
+
     #include <fcntl.h>
     #include <unistd.h>
     #include <sys/wait.h>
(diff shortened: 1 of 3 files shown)


CompilerSource/general/bettersystem.cpp
```diff
@@ -252,20 +256,90 @@ void myReplace(std::string& str, const std::string& oldStr, const std::string& n
...
     #include <crt_externs.h>
     #define environ (*_NSGetEnviron())
     extern char **environ;
 #endif
 
+    using std::vector;
+
+#if CURRENT_PLATFORM_ID == OS_FREEBSD
+    void ProcIdFromParentProcId(pid_t parentProcId, pid_t **procId, int *size) {
+      vector<pid_t> vec; int i = 0; int cntp;
```
JoshDreamland, line 285: Why not just return the vector instead of copying it to a new address?

CompilerSource/general/bettersystem.cpp
```diff
@@ -252,20 +256,90 @@ void myReplace(std::string& str, const std::string& oldStr, const std::string& n
...
+          vec.push_back(proc_info[j]); i++;
+        }
+      }
+      *procId = (pid_t *)malloc(sizeof(pid_t) * vec.size());
+      if (procId) {
+        std::copy(vec.begin(), vec.end(), *procId);
+        *size = i;
+      }
+    }
+#endif
```
JoshDreamland, line 327: You might as well just add an `#else` and implement `ProcIdFromParentProcId` as `waitpid(-fk,&result,__WALL);`. The `options` int seems to be zero for the other platforms, so you might consider dropping it rather than just ignoring it.

CompilerSource/general/bettersystem.cpp
time-killer-games: Sorry, old habit. I prefer doing C abi usually for my work in case i want to port the function to a shared library, which std::vector is implementation defined and won't work across different compilers for that. Anyway, I fixed it.

CompilerSource/general/bettersystem.cpp
time-killer-games: Good idea, just did it.

JoshDreamland (approved):

PR #2204, opened 2021-03-25
Cifixes

fundies:
Moves Android job to azure. Removes all build flags added for old Ubuntu. No more CI hacks


JoshDreamland (approved):
Wow, that was a lot of cruft.

PR #2214, opened 2021-05-04
libwebm

time-killer-games:
To test this pr, (only works on FreeBSD so far, will correct that soon):
*install the usual deps from the wiki, then run:*
```
git clone --recurse-submodules -b patch-75 https://github.com/time-killer-games/enigma-dev-fork ~/enigma-dev
cd ~/enigma-dev/
./install.sh
make
make emake
cd ~/enigma-dev/shared/libwebm/
cmake .
make
cd ../../
java -jar lateralgm.jar
```
Open this gm81 in LGM and run it: [WebMPlayer.zip](https://github.com/enigma-dev/enigma-dev/files/6434768/WebMPlayer.zip)

Needs changes to "Universal_System/bufferstruct.cpp" from https://github.com/enigma-dev/enigma-dev/pull/2215 to get the visual part of the video working (so you don't get a black screen with nothing but vorbis audio playback).

Issue #567, opened 2013-12-11
Buffers/Database Management


fundies:
Data management is an important aspect of many games. Strategy games, such as Age of Empires, often use databases as a means of storing character and building statistics. RPG games may use databases for similar reasons to cache weapon statistics and damage characteristics. ENIGMA lacks good data management facilities outside of its rudimentary data structure support. The advantage of having a full SQL-enabled API would not only make it easier for novices to implement them into their games, it would make it easier for them to write _efficient_ CRUD routines without prior knowledge of data structures.

1) A contributor would start by finishing up and polishing the existing binary buffers API in the Universal System. They would provide coverage tests to ensure the binary buffer alignment, serialization, and deserialization work correctly.
2) They would move on to creating a new extension to the Universal System that adds support for SQL databases using an appropriate "light" dependency suitable for game engines, such as SQLite.
3) To finish up the contributor could research and implement another extension for NoSQL or graph-based databases.

time-killer-games:
lol polyz

Ahmed-nd:
Hi guys, I am Ahmed Nasser,  Computer science a 4th year student, And it's my pleasure to be here.
I have experience in C++, and I'm interested in this project, And I hope to get to be a part of the GSOC 2022 program, it's my first time. Could you please guild me?how to start?

PR #2215, opened 2021-05-08
Implement buffer_set_surface().

time-killer-games:
It works!!!!!

Allows my video player in https://github.com/enigma-dev/enigma-dev/pull/2214 to render pixels to a surface from buffer_get_address().

Screenshot_20210508_215944

Screenshot taken on FreeBSD. Note FreeBSD is the only platoform my other pr (the video player one) currently works in, on Windows and Linux you will get undefined references, which I am in the process of figuring out before that pr will be ready for merging. 

However this is a separate task entirely, and should be good to go when the CI passes.

### DIFF
--- a/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
@@ -269,7 +269,6 @@ void buffer_get_surface(int buffer, int surface, int mode, unsigned offset, int 
 }
 
 void buffer_set_surface(int buffer, int surface, int mode, unsigned offset, int modulo) {
-  get_buffer(binbuff, buffer);
   int tex = surface_get_texture(surface);
   int wid = surface_get_width(surface);
   int hgt = surface_get_height(surface);
--- a/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
@@ -21,6 +21,7 @@
 
 #include "Resources/AssetArray.h" // TODO: start actually using for this resource
 #include "Graphics_Systems/graphics_mandatory.h"
+#include "Graphics_Systems/General/GSsurface.h"
 #include "Widget_Systems/widgets_mandatory.h"
 
 #include <cstring>
@@ -261,18 +262,18 @@ int buffer_get_type(int buffer) {
   return binbuff->type;
 }
 
-//NOTE: This function should most likely be added in graphics systems.
 void buffer_get_surface(int buffer, int surface, int mode, unsigned offset, int modulo) {
   //get_buffer(binbuff, buffer);
   //TODO: Write this function
   DEBUG_MESSAGE("Function unimplemented: buffer_get_surface", MESSAGE_TYPE::M_WARNING);
 }
 
-//NOTE: This function should most likely be added in graphics systems.
 void buffer_set_surface(int buffer, int surface, int mode, unsigned offset, int modulo) {
-  //get_buffer(binbuff, buffer);
-  //TODO: Write this function
-  DEBUG_MESSAGE("Function unimplemented: buffer_set_surface", MESSAGE_TYPE::M_WARNING);
+  get_buffer(binbuff, buffer);
+  int tex = surface_get_texture(surface);
+  int wid = surface_get_width(surface);
+  int hgt = surface_get_height(surface);
+  enigma::graphics_push_texture_pixels(tex, wid, hgt, (unsigned char *)buffer_get_address(buffer));
 }
 
 void buffer_resize(int buffer, unsigned size) {
--- a/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
@@ -275,7 +275,7 @@ void buffer_set_surface(int buffer, int surface, int mode, unsigned offset, int 
   if (buffer_get_size(buffer) == buffer_sizeof(buffer_u64) * wid * hgt) {
     enigma::graphics_push_texture_pixels(tex, wid, hgt, (unsigned char *)buffer_get_address(buffer));
   } else { // execution can not continue safely with wrong buffer size
-    DEBUG_MESSAGE("Buffer allocated with wrong length! Cannot continue...", MESSAGE_TYPE::M_FATAL_ERROR);
+    DEBUG_MESSAGE("Buffer allocated with wrong length!", MESSAGE_TYPE::M_WARNING);
   }
 }
 
--- a/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
+++ b/ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
@@ -272,7 +272,11 @@ void buffer_set_surface(int buffer, int surface, int mode, unsigned offset, int 
   int tex = surface_get_texture(surface);
   int wid = surface_get_width(surface);
   int hgt = surface_get_height(surface);
-  enigma::graphics_push_texture_pixels(tex, wid, hgt, (unsigned char *)buffer_get_address(buffer));
+  if (buffer_get_size(buffer) == buffer_sizeof(buffer_u64) * wid * hgt) {
+    enigma::graphics_push_texture_pixels(tex, wid, hgt, (unsigned char *)buffer_get_address(buffer));
+  } else { // execution can not continue safely with wrong buffer size
+    DEBUG_MESSAGE("Buffer allocated with wrong length! Cannot continue...", MESSAGE_TYPE::M_FATAL_ERROR);
+  }
 }
 
 void buffer_resize(int buffer, unsigned size) {


ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
```diff
@@ -261,18 +262,18 @@ int buffer_get_type(int buffer) {
...
   //TODO: Write this function
   DEBUG_MESSAGE("Function unimplemented: buffer_get_surface", MESSAGE_TYPE::M_WARNING);
 }
 
-//NOTE: This function should most likely be added in graphics systems.
 void buffer_set_surface(int buffer, int surface, int mode, unsigned offset, int modulo) {
-  //get_buffer(binbuff, buffer);
-  //TODO: Write this function
-  DEBUG_MESSAGE("Function unimplemented: buffer_set_surface", MESSAGE_TYPE::M_WARNING);
+  get_buffer(binbuff, buffer);
```
JoshDreamland, line 272: The `mode`, `offset`, and `modulo` parameters are unused. I'm not clear on how the results of `get_buffer` are being used, here; seems you're just using `buffer_get_address`.

ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
```diff
@@ -261,18 +262,18 @@ int buffer_get_type(int buffer) {
...
+  int tex = surface_get_texture(surface);
+  int wid = surface_get_width(surface);
+  int hgt = surface_get_height(surface);
+  enigma::graphics_push_texture_pixels(tex, wid, hgt, (unsigned char *)buffer_get_address(buffer));
```
JoshDreamland, line 276: What happens if `wid * hgt` is less than the buffer size? If it's substantially more, this will probably crash.

time-killer-games: Checking the buffer size for comparison now, but in the meantime it's worth noting the last two parameters of buffer_set_surface are no longer used as of GMS 2.3.x and are taken out as obsolete. https://manual.yoyogames.com/GameMaker_Language/GML_Reference/Buffers/buffer_set_surface.htm I used zero for the offset because that's what nik used as he helped write this code.

time-killer-games: surface_get_width(surf) * surface_get_height(surf) = 
Screenshot_20210508_232935
buffer_get_size(buff) = 
Screenshot_20210508_233034

time-killer-games: Also try loading this file into lgm https://github.com/enigma-dev/enigma-dev/files/6434768/WebMPlayer.zip

It might give you a better idea of how I'm using the function.

JoshDreamland: I'm sure it doesn't crash when the size is right. But you don't have any checks to verify that the size is right.

time-killer-games: What do you mean by add a check? This isn't my video player or, I am just relying on the function this pr provides in that other one. If you want me to add a CI test or something, it would make more sense to add several to my other pr, I'm not so sure what test can be written that would be relevent to adding the one single function this pr adds. I'm confused by what you are asking.

time-killer-games: @JoshDreamland the number drawn on top of the video is the value produced by the code I wrote above each screenshot. Sorry I should've made it more obvious what I was doing with those screenshots.

time-killer-games: @JoshDreamland so you're saying you can't see the checks in those screenshots and additional background information I provided for them?

time-killer-games: I really don't know what more you want me to do, I literally did `draw_text(x,y,string(surface_get_width(surf) * surface_get_height(surf)))` and `draw_text(x,y,string(buffer_get_size(buff)))` and then did the math on those numbers and found out 1843200 = 8 * 230400 and 320400 = w * h and 1843200 = buffer_get_size().

JoshDreamland: I'm not saying the code doesn't work, or that you aren't verifying that the size is right in your GML. I'm saying that if a user *didn't* do that properly in *their* GML, this would behave badly or crash, because this routine that I am reviewing here has no such checks.

time-killer-games: Pixel sizes can vary depending on the format, target architecture, whether there is any alpha channel, and other things, so it seems nearly impossible for me to predict off-hand whether a particular user is going to use the right buffer size before hand and prevent them from using wrong sizes if they attempt to. I'm not a fortune teller. You still haven't told me what you even want me to do for these magical checks, or what is even being checked or how. Just about the only thing that remains, certain is the number of pixels being read from the width/height, however the size of each pixel is up in the air as far as I can tell.

time-killer-games: Tell me what you are proposing for the first time and i'll do it. You are avoiding to answer my question.

What do you mean by "check" to begin with? 

i am not aware of any means to ensure beforehand the user doesn't use an incorrect buffer size. 

YoYoGames will allow the user to crash the game, what's so terrible if we did the same?

Pixel size will vary depending on a lot of things, from target architecture, to whether there is an alpha channel, and more. Is it safe to assume each pixel is exactly 8 bits in every scenerio? Doesn't seem like that would be the case, but I don't know enough about this to be the one to say.

It would make my life a lot easier if I didn't have to fight tooth an nail for you to spit out what you mean.

JoshDreamland: I assumed you knew what I mean based on the fact that you were setting the correct length in your GML. That is the one and only correct length. You're right, there are multiple pixel formats—but you're not passing in a pixel format. The buffer you're passing doesn't contain a pixel format. Think this through—how is `graphics_push_texture_pixels` supposed to know what format your pixel data has? You aren't passing a buffer length. You're only passing a width and a height.

The buffer length you're using in your GML is the only correct length for that width and height. I'm just asking you to check that the buffer actually has that length before sending it to the graphics system to read. Otherwise, it will either read too few pixels and look wrong, or too many pixels and crash.

time-killer-games: as we had previously went over, buffer_get_size() = buffer_sizeof(buffer_u64) * surface_get_width(surf) * surface_get_height(surf). buffer_get_size() is the buffer length, no? buffer_sizeof() is the size of each pixel, which is the third part of required info we need to calculate the buffer length, the other two being the width and height. buffer_sizeof won't always have buffer_u64 passed to it as an argument, so again I ask, how to go about this check? There seems to be no other way to do it other than predict the pixel format.

fundies: you've shown checking it in your GML just check it inside your c++ function and throw a debug msg if sizes don't match up.

time-killer-games: Honestly, I'm not trying to be difficult, but i just looked at my gm81 project again and don't see what you are referring to. Sorry guys, I just might be braindead. :(

JoshDreamland: I think what you're missing is just that there isn't "guessing" the pixel format at play—there can only be one supported pixel format unless you were specifying the format by some other means. The format isn't an argument to the function (`graphics_push_texture_pixels`), and it's not stored in the pixel buffer you pass (nor is the actual length of the pixel buffer), so the format you're passing is the only one supported.

RobertBColton: We're still missing grow buffers. With a grow buffer that should be no problem. Can worry about that later, nice work adding this feature! #567 is open and available if you want to continue the work on buffers.

ENIGMAsystem/SHELL/Universal_System/bufferstruct.cpp
```diff
@@ -261,18 +262,21 @@ int buffer_get_type(int buffer) {
...
-  //get_buffer(binbuff, buffer);
-  //TODO: Write this function
-  DEBUG_MESSAGE("Function unimplemented: buffer_set_surface", MESSAGE_TYPE::M_WARNING);
+  int tex = surface_get_texture(surface);
+  int wid = surface_get_width(surface);
+  int hgt = surface_get_height(surface);
+  if (buffer_get_size(buffer) == buffer_sizeof(buffer_u64) * wid * hgt) {
+    enigma::graphics_push_texture_pixels(tex, wid, hgt, (unsigned char *)buffer_get_address(buffer));
+  } else { // execution can not continue safely with wrong buffer size
+    DEBUG_MESSAGE("Buffer allocated with wrong length! Cannot continue...", MESSAGE_TYPE::M_FATAL_ERROR);
```
RobertBColton, line 278: Is this really a fatal error?

time-killer-games: nah; wait a few seconds while I fix that.

JoshDreamland (approved):
Looking nice and safe, now; thanks!

RobertBColton:
You can implement `buffer_get_surface` later using `graphics_copy_texture_pixels`, there's a whole lot of low-level DMA mandatory texture functions I added to solve these problems.

PR #2216, opened 2021-05-12
make mode optional, make offset not be

time-killer-games:
I did it wrong in my previously merged pr. Since GMS 2.3.x, mode and modulo are omitted, offset is not. my mistake.

https://manual.yoyogames.com/GameMaker_Language/GML_Reference/Buffers/buffer_get_surface.htm
https://manual.yoyogames.com/GameMaker_Language/GML_Reference/Buffers/buffer_set_surface.htm


fundies:
Only the last x amount of arguments can have default values you cannot mix and match like that. If you want to omit the arguments in that order you will need to have multiple declarations of the function

fundies:
you also need to implement it by having it call the longer version...

time-killer-games:
> you also need to implement it by having it call the longer version...

I'm waiting for my other pr to get merged first so i can resolve conflicts that will arise.

PR #2217, opened 2021-05-16
More macOS Compile Error Fixes

time-killer-games:
Fixes include:
- Objective C has no concept of namespaces so I had to move cocoa_window_handle() from an objective c source to a cpp source. For the same reason I can't put that function in a namespace or it won't be accessible to the Obj-C side.
- Undefined reference to remove_trailing_zeros() solved.
- Typo with windowNumber fixed.
- Forgot a necessary type cast.


time-killer-games:
@JoshDreamland what do i do? OS X needs those widget declarations static in those headers but we have no means to do that without breaking all our other platforms?

fundies:
> @JoshDreamland what do i do? OS X needs those widget declarations static in those headers but we have no means to do that without breaking all our other platforms?

It needs what to be static and why? If you can't call debug_message because it's in a namespace then you can wrap it in a static function for cocoa. But i'm not sure what your actual problem is.

time-killer-games:
> > @JoshDreamland what do i do? OS X needs those widget declarations static in those headers but we have no means to do that without breaking all our other platforms?
> 
> It needs what to be static and why? If you can't call debug_message because it's in a namespace then you can wrap it in a static function for cocoa. But i'm not sure what your actual problem is.

literally just about all of our widget functions are causing clang to cry about duplicate definitions for architecture x86_64 and the only known means for me to silence those errors so far is making all of our widget functions static in the headers the declare them. I only managed to find this "fix" on stackoverflow but I know it's not correct otherwise it wouldn't break all our other platforms. There should be a way to do this that "just works".

Unfotunately, contrary to popular belief, Apple products don't "just work" or we wouldn't have issues like this.

fundies:
> > > @JoshDreamland what do i do? OS X needs those widget declarations static in those headers but we have no means to do that without breaking all our other platforms?
> > 
> > 
> > It needs what to be static and why? If you can't call debug_message because it's in a namespace then you can wrap it in a static function for cocoa. But i'm not sure what your actual problem is.
> 
> Literally just about all of our widget functions are causing clang to cry about 'duplicate definitions for architecture x86_64' and the only known means for me to silence those errors so far is making all of our widget functions static in the headers which declare them. I only managed to find this "fix" on stackoverflow but I know it's not correct otherwise it wouldn't break all our other platforms. There should be a way to do this that "just works".
> 
> Unfotunately, contrary to popular belief, Apple products don't "just work" or we wouldn't have issues like this. The thing is, there literally are no duplicate definitions that I can see, and we have the same amount of definitions for each function as we do all our other platforms: one for each function. Really weird stuff.

OSX prolly has these defined in cocoa platform and the none widget system. That cocoa code is way out of date.

time-killer-games:
> > > > @JoshDreamland what do i do? OS X needs those widget declarations static in those headers but we have no means to do that without breaking all our other platforms?
> > > 
> > > 
> > > It needs what to be static and why? If you can't call debug_message because it's in a namespace then you can wrap it in a static function for cocoa. But i'm not sure what your actual problem is.
> > 
> > 
> > Literally just about all of our widget functions are causing clang to cry about 'duplicate definitions for architecture x86_64' and the only known means for me to silence those errors so far is making all of our widget functions static in the headers which declare them. I only managed to find this "fix" on stackoverflow but I know it's not correct otherwise it wouldn't break all our other platforms. There should be a way to do this that "just works".
> > Unfotunately, contrary to popular belief, Apple products don't "just work" or we wouldn't have issues like this. The thing is, there literally are no duplicate definitions that I can see, and we have the same amount of definitions for each function as we do all our other platforms: one for each function. Really weird stuff.
> 
> OSX prolly has these defined in cocoa platform and the none widget system. That cocoa code is way out of date.

Thanks, I'll check that out. However, I'm building with SDL, not cocoa, what you say may still be related, I'm not sure. I am using Cocoa Widgets, but not the actual Cocoa platform.

time-killer-games:
@fundies @JoshDreamland I can't believe I am saying this but I think I got macOS building games in LGM again finally. emake is also close to working.

time-killer-games:
Screen Shot 2021-05-21 at 11 47 53 PM

fundies:
Cool I don't think osx can read data on off the exe so in the.ey file you will need to set the $exe bar to an actual file

fundies:
pretty sure you just removed the wrong one. did you test this locally?

time-killer-games:
The screenshot would imply that I at least used to have it in a compiling state. The way I interpreted what you said I wasn't sure which one needed to be removed.

fundies:
> The screenshot would imply that I at least used to have it in a compiling state. The way I interpreted what you said I wasn't sure which one needed to be removed. I must've skimmed past the "remove the first one" XD

Well I gave yo what I think is causing the error to try. No need to commit if it doesn't work locally

time-killer-games:
Screen Shot 2021-05-23 at 8 46 14 PM

Here's the app I built, tell hugar to get his MacBook back from the islanders overseas and test it
[xProcess.zip](https://github.com/enigma-dev/enigma-dev/files/6529203/xProcess.zip)

fundies:
> Screen Shot 2021-05-23 at 8 46 14 PM
> 
> Here's the app I built, tell hugar to get his MacBook back from the islanders overseas and test it
> [xProcess.zip](https://github.com/enigma-dev/enigma-dev/files/6529203/xProcess.zip)

Still looks borked. I have some ideas of things to try tho if you want to msg me on discord

CompilerSource/general/bettersystem.cpp
```diff
@@ -310,7 +310,7 @@ void myReplace(std::string& str, const std::string& oldStr, const std::string& n
       proc_listpids(PROC_ALL_PIDS, 0, &proc_info[0], sizeof(pid_t) * cntp);
       for (int j = cntp; j > 0; j--) {
         if (proc_info[j] == 0) { continue; }
-        if (ppid == ParentProcIdFromProcId(proc_info[j])) {
+        if (parentProcId == ParentProcIdFromProcId(proc_info[j])) {
```
time-killer-games, line 313: @JoshDreamland this would be important to get mac working in bettersystem

JoshDreamland (approved):
Looks good to me; thanks for giving Mac some attention!

PR #2221, opened 2021-05-23
Add function window_get_handle().

time-killer-games:
The reverse of window_get_identifier().

This function will be used in an upcoming pr.


ENIGMAsystem/SHELL/Platforms/Win32/WINDOWSwindow.cpp
```diff
@@ -39,6 +39,8 @@
 #endif
 
 #include <stdio.h>
```
fundies, line 41: if you include cstdio you shouldn't need stdio.h anymore. the "<cheader>" versions are just namespaced I think

time-killer-games: nice catch! I didn't actually notice the duplicate header at first.

Issue #2203, opened 2021-03-25
C++17 temp_directory_path() broken on Windows


time-killer-games:
I can tell Josh and fundies have a fetish for having broken c++17 features in enigma. emake broken.

https://github.com/enigma-dev/enigma-dev/blob/792badac3b619b1c81ce1d106b813e83ebcb5aa9/CommandLine/emake/Main.cpp#L32

Also please use path.u8string() instead of path.string() to avoid creating more tech debt for when this code eventually works on english-only machines.

If you guys ever asked me to do code review, pitiful mistakes like these wouldn't be merged, because I actively use and test all our desktop platforms and value them as equals and not with a Linux bias/fetish.

fundies:
broken how? Is it crashing? Always or only if you pass a gmk? Also, you can review any PR you want they're all public

time-killer-games:
Running emake complains with an error of temp_directory_path failing because the c++17 devs never got it working on Windows. Last I checked you already knew this because we talked about it before and you confirmed it with Robert or whoever so it needs no further explanation

fundies:
It's just a log file you shouldn't need it to work though

time-killer-games:
Yes but the fact it's even called prevents running emake. I believe this is probably due the fact you didn't use the no except version but I'll double check the docs

fundies:
Also afaict the CI is running empty games on windows fine. If it's just running projects that broke you can add a test for that pretty easy but I'd rather someone port the test harness to windows

time-killer-games:
It could be I may need a package update and they may have fixed it. I'll install master to verify. I was too quick to delete my install.

time-killer-games:
```
#include <iostream>
#include <filesystem>

int main() {
  std::cout << std::filesystem::temp_directory_path().string() << std::endl;
  return 0;
}
```

![Capture](https://user-images.githubusercontent.com/4379204/112532079-c4ad0080-8d7e-11eb-9443-8b55a09a9375.PNG)

time-killer-games:
where as:

```
#include <iostream>
#include <filesystem>
#include <system_error>

int main() {
  std::error_code ec;
  std::cout << std::filesystem::temp_directory_path(ec).string() << std::endl;
  return 0;
}
```

![Capture](https://user-images.githubusercontent.com/4379204/112532623-6cc2c980-8d7f-11eb-8a45-5652080d33fa.PNG)

[...]

PR #2225, opened 2021-05-29
std::filesystem::temp_directory_path() broken fix for Windows. (MinGW 32-bit)

time-killer-games:
Mostly fixes https://github.com/enigma-dev/enigma-dev/issues/2203. see screenshots and code snippets in that issue for more details, evidence, reproducible, etc. The only thing I didn't address was adding UTF-8 support. @fundies never removed the unconfirmed tag because he put his brain in the blender.

The bug only happens with MinGW 32-bit. It works just fine when built for 64-bit. Bug not present in MSVC.


time-killer-games:
@JoshDreamland merge pls

PR #2232, opened 2021-06-10
decouple none [wip]

fundies:


ENIGMAsystem/SHELL/Platforms/None/Makefile
```diff
@@ -1,7 +1,2 @@
-ifeq ($(OS), Linux)
-	SOURCES += $(wildcard Platforms/General/POSIX/Linux/*.cpp)
-else ifeq ($(OS), FreeBSD)
-	SOURCES += $(wildcard Platforms/General/POSIX/BSD/*.cpp)
-endif
-SOURCES += $(wildcard Platforms/None/*.cpp) $(wildcard Platforms/General/POSIX/*.cpp) Platforms/General/PFshell/PFshell.cpp
```
time-killer-games, line 6: This platform was originally added by faissaloo with the intent for it to be used for headless servers. Thus, this platform completely detached from it's original purpose and turned into something utterly useless that no one would ever use without access to the shell. Consider moving POSIXshell and WINDOWSshell to an extension so that the user may still use them with the None platform without them being mandatory (although the None platform is 100% useless without it, there is no same reason for it to exist without such a feature as headless servers are all it is even useful for in any way.

time-killer-games: @fundies what I said was a mere suggestion. I am happy to do that in a follow up pr to this one if you aren't up to it, just let me know what to expect and I'll act accordingly. :)

fundies: Me or someone else needs to update our delta time stuff to be cross compatible before I merge this so don't panic yet

time-killer-games: @fundies we're on a baby hunt! and don't think we don't know how to weed em' out!

PR #2235, opened 2021-06-17
Add xProcess Library Functions to Shared

time-killer-games:
This would really help better system look a lot cleaner and easier to read. 

```
// execute command, return process id of running executable instance specified
PROCESS ProcessExecuteAsync(const char *command);
```

```
// write string to standard input for the specified process id (if ran with previous function)
void ExecutedProcessWriteToStandardInput(PROCESS procIndex, const char *input)
```

```
// read string from standard input of process id (if ran from ProcessExecuteAsync)
const char *ExecutedProcessReadFromStandardOutput(PROCESS procIndex);
```

```
// free strings and erase maps from memory when done 
void FreeExecutedProcessStandardInput(PROCESS procIndex);
void FreeExecutedProcessStandardOutput(PROCESS procIndex);
```

```
// check if a process previously ran with ProcessExecuteAsync has closed
bool CompletionStatusFromExecutedProcess(PROCESS procIndex);
```

```
// these can be used to get the window id from process id for CI tests
PROCINFO ProcInfoFromProcId(PROCID procId);
WINDOWID OwnedWindowId(PROCINFO procInfo, int i);
int OwnedWindowIdLength(PROCINFO procInfo);
void FreeProcInfo(PROCINFO procInfo);

```
These functions as well as others can also be used in the engine in various places.

Supports UTF-8 with all functions.


JoshDreamland:
Where's the primary repository for this? How do we pull the new version when this inevitably stops building?

time-killer-games:
> Where's the primary repository for this? How do we pull the new version when this inevitably stops building?

That would be here: https://github.com/time-killer-games/xProcess

Feel free to fork it for (not to actually contribute (as that I wouldn't impose on you the need to do that as long as I'm around) but for backup purposes I.e. if I get banned from gihub again for some reason or I disappear for some reason you'll have something to fall back on.

I may also provide extensive code comments or remove code you find we'll never have the need for, but for the sake of ease of mantanance, it's better to keep things 1:1 with my repository so I don't need to constantly maintain two separate versions like I already have to with my dialogs and other various things.

time-killer-games:
In regards to supporting OS X CI in the future, the Window ID type provided by this code is a CGWindowID and can be used with code pulled from Apple's 'son of grab' example project to take automated screenshots for regression tests like we do on our other platforms: https://stackoverflow.com/questions/24321222/how-to-build-apples-son-of-grab-example

time-killer-games:
This project will eventually need to support Wayland and I intend to add that well before we even support that as a platform in the engine itself as I was planning to look into that relatively soon actually. For headless games with None for graphics and/or platform this may be build without the -DXPROCESS_GUIWINDOW_IMPL definition flag and it'll omit windowing for users who might want to implement the code in an extension for example, I'd like to replace our POSIXshell and WINDOWSshell sources with an extension which simply does the same thing they always did but wraps this library.

time-killer-games:
I intend to provide docs on all the features to ensure you know what everything this lib does and can make the most of it.

JoshDreamland:
I think that would not be worth your effort—this library may only be half the size as Boost Process, but it's not as full-featured, doesn't have comparable testing or usership, and we'd be (or you'd be) directly responsible for fixing breakages and adding new features. We should adopt Boost Process before writing up a similar library.

time-killer-games:
@JoshDreamland I'd actually like working with this, and this has a good number of features the boost process library does not have. Their library can only get the environment, cmdline, and cuirrent working directory of the current process, where as this can do that with any process running on your user session so long as you know what the process id is of the process you need to inspect, which are great utility functions to have for debugging purposes among other things. Not to mention this fills the gap between processes in windowing when built with the proper compiler flags, which boost also doesn't cover afaict.

The fact is, a lot of these features we already have implemented on our own in master, this would provide all those same features but with code that is less messy and wasn't done in a half-assed manner. I am very familiar with most if not all API's used here and this code is very reliable and won't break for the foreseeable future. I'd appreciate it if you could toss your arrogance and lack of faith in me aside and out the window it's the only thing keeping us from making progress on my end.

time-killer-games:
@JoshDreamland what separates me from you in another aspect, when I call you an insult, I do not mean it, and often regret it, I say it usually because you did something provoking and childish which prompted and called for correction. Where as when you insult me, you really believe the garbage you say to me, 100% of the time, and it isn't a matter of being upset with me, it's just how you choose to feel about me on an ongoing basis regardless of your current mood and without regrets.

That's what you would call an extremely negative person. The kind of person I typically avoid in real life. It's more difficult to leave this project now though because of how many years I have been working with you on it. I'd appreciate more respect, less insults, and less arguing, real dialog instead of just discarding all my work i spend countless hours on and without any reasoning based on experience attached to it, you flat out assume this code will constantly break going no real basis.

PR #2248, opened 2021-07-30
Add xlib-None Graphics Bridge.

time-killer-games:


ENIGMAsystem/SHELL/Platforms/xlib/XLIBwindow.cpp
```diff
@@ -113,7 +119,15 @@ bool initGameWindow()
   if (winh > screen->height) winh = screen->height;
 
   // Make the window
+  #if !defined(GRAPHICS_NONE)
```
fundies, line 122: These macros are going to cause issues on system changes you need to conditionally build them and stubs using the makefiles instead

time-killer-games: I've done what you said locally. I'll create a new pull request and close this one.

PR #2266, opened 2021-08-25
bash is bad

time-killer-games:
Not everyone likes being forced to install software just to run two small shell scripts that if written better they wouldn't need it. Not only that, Bourne shell is always located at /bin/sh. Bash is not always located at /bin/bash depending on the OS.

The install script for the jars was also an unnecessary amount of lines and written in such a way that it was over-complicating things that should be extremely simple.

MacOS is moving in the direction of removing bash as per the license change to GPL, this is why they changed zsh to be the default shell. Mac will always have /bin/sh. I mention Mac and not the other two platforms this fixes major issues with, because I know you guys don't really care about my BSD fetish; I understand your reasons.

I'm anticipating one of you guys telling me to add back the ".deps" and "checking if the files are already updated" shell logic, but I'm not sure how to do that in a way that is Bourne shell compliant as a lot of this is relatively new to me. Examples of how to do that are welcome. Otherwise, this pr will take a lot longer than what is necessary.

I allowed edits from the maintainers, and that's only with the assumption you don't accidentally or intentionally change it back to relying on bash-specific script logic. When this gets merged I'm going to see about updating the wiki install scripts and gut out the use of bash. Whoever found that necessary initially remains a mystery.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,8 +1,9 @@
-#!/bin/bash
+#!/bin/sh
 
-declare -A deps
-deps["lateralgm.jar"]="IsmAvatar/LateralGM"
-deps["plugins/enigma.jar"]="enigma-dev/lgmplugin"
+key0="enigma-dev/lgmplugin"
+key1="IsmAvatar/LateralGM"
+dep0="plugins/enigma.jar"
+dep1="lateralgm.jar"
 
 get_latest () {
   local release=$(curl --silent "https://api.github.com/repos/$1/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -27,31 +28,11 @@ else
...
-fi
-  
-lineCount=$(wc -l < ".deps")
-lineNum=1;
-
-for key in ${!deps[@]}; 
-do
-  
-  latest=$(get_latest ${deps[$key]})
+download_latest() {
```
JoshDreamland, line 31: Accept key/value here, then list out `download_latest enigma-dev/lgmplugin IsmAvatar/LateralGM`, etc, below

fundies:
Sounds more like MacOS is bad but whatever

fundies (changes requested):
Fix the deps file and I don't mind if this merged but I think it's unnecessary overall

PR #2278, opened 2021-10-11
pass represents / depends through grpc

fundies:


CommandLine/emake/Server.cpp
```diff
@@ -127,6 +127,28 @@ class CompilerServiceImpl final : public Compiler::Service {
         if (author.empty())
           author = about.get("maintainer");
 
+        eyit represents = about.values.find("represents");
+        if (represents != about.values.end()) {
+          std::string repsStr = (represents->second)->data().get("build-platforms");
```
JoshDreamland, line 132: I don't remember if `data()` returns a sentinel. Might be worth checking that it does.

JoshDreamland (approved):

PR #2293, opened 2022-03-29
Fix 'folder replaces resource' egm-read issue

KartikShrivastava:
This is a small fix for the issue in which on loading egm projects in RGM, folders appears in place of actual resources.

The issue is a side effect of calling `mutable_folder` before checking if node contains `folder` or not, as a call to `mutable_folder` creates an unwanted `folder` which was happening in `void RecursiveResourceSanityCheck(...)` of `egm-read.cpp`.


KartikShrivastava:
Hi @fundies,
I've tested reading egm which was saved with nested folder structure.

image

JoshDreamland (approved):
Thanks for patching this!

PR #2295, opened 2022-04-17
Fix more build errors with JDI update

dc03:
This fixes the build errors which showed up when running `make all`

NOTE: The changes in `jdi_utility.cpp` should be reviewed, as they are just a hack to get the build to finish

### DIFF
--- a/CompilerSource/compiler/jdi_utility.cpp
+++ b/CompilerSource/compiler/jdi_utility.cpp
@@ -132,6 +132,8 @@ bool lang_CPP::global_exists(string n) const {
 void lang_CPP::quickmember_variable(jdi::definition_scope* scope, jdi::definition* type, string name) {
   scope->members[name] = std::make_unique<jdi::definition_typed>(name,scope,type);
 }
+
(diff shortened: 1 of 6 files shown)


CompilerSource/compiler/jdi_utility.cpp
```diff
@@ -132,6 +132,8 @@ bool lang_CPP::global_exists(string n) const {
 void lang_CPP::quickmember_variable(jdi::definition_scope* scope, jdi::definition* type, string name) {
   scope->members[name] = std::make_unique<jdi::definition_typed>(name,scope,type);
 }
+
+enigma::parsing::StdErrorHandler hackybaby;  // TODO: FIXME: This should be using a central error handler...
```
JoshDreamland, line 136: More or less exactly what we do elsewhere (e.g. `lang_CPP.cpp`), except we keep it in the function instead of making it a global.

```cpp
   enigma::parsing::StdErrorHandler hack;  // TODO: FIXME: This should be using a central error handler...
```

JoshDreamland (approved):
Thanks for fixing these! I think this successfully migrates us to the new JDI.

PR #2299, opened 2022-05-30
Update Makefile to use PKG-CONFIG

HitCoder9768:


ENIGMAsystem/SHELL/Audio_Systems/OpenAL/Makefile
```diff
@@ -1,3 +1,7 @@
+ifndef PKG-CONFIG
```
fundies, line 1: I think this belongs in https://github.com/enigma-dev/enigma-dev/blob/master/Config.mk and the SDL one can be deleted but needs testing on CI

PR #2305, opened 2022-06-10
WIP: Fixes for macros in EDL lexer

dc03:
These series of commits:
1. Adds a test to check macro expansion
2. Fixes the closing parenthesis of macros being considered as a macro argument
3. Fixes incorrect strings being used when stringifying a macro argument

### DIFF
--- a/CompilerSource/parsing/lexer.cpp
+++ b/CompilerSource/parsing/lexer.cpp
@@ -432,7 +432,7 @@ bool Lexer::HandleMacro(std::string_view name) {
         for (size_t i = 0; i < args.size(); ++i)
           args_evald[i] = PreprocessBuffer(args[i]);
         PushMacro(macro.name,
-                  macro.SubstituteAndUnroll(args, args_evald, errc));
+                  macro.SubstituteAndUnroll(args, args_evald, errc, stringified_macros));
       } else {
         errc.Error() << "No closing parenthesis for arguments to `"
                      << name << '`';
--- a/CompilerSource/parsing/lexer.cpp
+++ b/CompilerSource/parsing/lexer.cpp
@@ -422,7 +422,7 @@ bool Lexer::HandleMacro(std::string_view name) {
         else if (t.type == TT_ENDPARENTH) --paren;
         if (t.type == TT_COMMA && paren == 1) {
           args.emplace_back();
-        } else {
+        } else if (paren != 0) {
           args.back().push_back(t);
         }
       } while (paren && t.type != TT_ENDOFCODE);
--- a/CompilerSource/parsing/lexer.h
+++ b/CompilerSource/parsing/lexer.h
@@ -127,6 +127,17 @@ class Lexer {
   const ParseContext *context;
   std::deque<OpenMacro> open_macros;
   ErrorHandler *herr;
+
+  /// Store the stringified versions of macros in a set so that they can be safely referred to by `std::string_view`.
+  ///
+  /// As `CodeSnippet`s point inside the source itself, it is not possible to make one that points to the `std::string`
+  /// created on stringifying a macro as that string doesn't exist within the source and can only be referred to within
+  /// the function it is created in and therefore when trying to access this string outside, it can cause a segfault.
+  ///
+  /// Thus, this set "extends" the lifetime of the stringified forms of macros by promoting their lifetimes to the
+  /// lifetime of the lexer itself. The reason a `std::unordered_set` is used is to deduplicate macros which stringify
+  /// to the same string.
+  Macro::StringifiedSet stringified_macros;
   
   struct Options {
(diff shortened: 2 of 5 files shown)


CompilerSource/parsing/lexer.cpp
```diff
@@ -422,7 +422,7 @@ bool Lexer::HandleMacro(std::string_view name) {
         else if (t.type == TT_ENDPARENTH) --paren;
         if (t.type == TT_COMMA && paren == 1) {
           args.emplace_back();
-        } else {
+        } else if (paren != 0) {
```
JoshDreamland, line 425: This is where the variadic check should be done, too... I just forgot. 😜

CompilerSource/parsing/lexer.h
```diff
@@ -127,6 +127,17 @@ class Lexer {
...
+  /// Store the stringified versions of macros in a set so that they can be safely referred to by `std::string_view`.
+  ///
+  /// As `CodeSnippet`s point inside the source itself, it is not possible to make one that points to the `std::string`
+  /// created on stringifying a macro as that string doesn't exist within the source and can only be referred to within
+  /// the function it is created in and therefore when trying to access this string outside, it can cause a segfault.
+  ///
+  /// Thus, this set "extends" the lifetime of the stringified forms of macros by promoting their lifetimes to the
+  /// lifetime of the lexer itself. The reason a `std::unordered_set` is used is to deduplicate macros which stringify
+  /// to the same string.
+  Macro::StringifiedSet stringified_macros;
```
JoshDreamland, line 140: Good call! Now, another option (that I do in JDI) is to let tokens optionally own their own content string. In JDI, a token owns strings like these, and also takes ownership of its content on copy (notably, when copying to a token buffer).

CompilerSource/parsing/macros.cpp
```diff
@@ -257,7 +256,12 @@ TokenVector Macro::SubstituteAndUnroll(
           for (const Token &tok : args[ind])
             str += tok.content;
           str = QuoteString(str);
-          TokenVector vec{Token(TT_STRINGLIT, errc.snippet)};
+
+          stringified_macros.insert(str);
```
JoshDreamland, line 260: This looks like a good place to avoid a second lookup by using the iterator returned by `insert`.

JoshDreamland (commented):
Great work! A few comments for your consideration...

CompilerSource/parsing/lexer.cpp
dc03: Well, I feel variadic-ness should be handled in `SubstituteAndUnroll` because this function just seems to handle reading in n-number of arguments but not validating any of them.

CompilerSource/parsing/macros.cpp
dc03: Fixed with c1181c90e9f7734aebaad288438c0ddddf13e44e! I was not aware that `insert` returned an iterator.

CompilerSource/parsing/lexer.h
dc03: One issue that comes to mind is that for almost all tokens in the source, you don't need to have them own their source. Thus, by adding the `mutable std::string owned_str;` to the `content` struct in JDI (which is what I assume stores the copied string), you're distributing the cost of storing the copied string across _all_ tokens in the source which means you end up using more memory as almost every token is now storing 24-32 (usually) extra bytes of an uninitialized `std::string`.

CompilerSource/parsing/lexer.cpp
JoshDreamland: But then we lose the comma tokens that separated these variadic arguments. To keep the original tokens, we need to have that context here.

CompilerSource/parsing/lexer.h
JoshDreamland: It's true. But memory is cheap, and so is the initialization. This way, every string we compose lives as long as the lexer, so we rely on the lexer being short-lived... but it still has to live as long as every token produced from it.

CompilerSource/parsing/lexer.cpp
dc03: I don't think that matters? As the macro arguments are stored as a `std::vector<TokenVector>`, each element of the vector contains all the tokens making up each argument, so it doesn't matter if commas are stored or not.

JoshDreamland: We can always spoof a new comma Token when we need to start returning the variadic argument tokens, sure, but that's extra processing we need to do on the spot. Usually, we never see the comma token again, but if your variadic args are "a, b", `__VA_ARGS__` needs to include that comma (and all the others). We lose the source information this way (e.g. the original code pointer), so if that comma turns out to add too many arguments to an underlying normal function, for example, the error reporting might get very confusing.

PR #2306, opened 2022-06-23
WIP: EDL parser implementation

dc03:
- Expression parsing is mostly done, type casting needs to be handled still, and functional casts are not treated specially
- Statement parsing needs to be implemented
- All the precedence declaration stuff has been moved into `precedence.h`


CompilerSource/parser/precedence.h
```diff
@@ -0,0 +1,118 @@
...
+  std::unordered_map<TokenType, OperatorPrecedence> Precedence::kUnaryPostfixPrec {
+      {TT_INCREMENT,    {Precedence::kUnaryPostfix, Associativity::LTR}},
+      {TT_DECREMENT,    {Precedence::kUnaryPostfix, Associativity::LTR}},
+  };
+
+  std::unordered_map<TokenType, OperatorPrecedence> Precedence::kTernaryPrec {
+      {TT_QMARK, {Precedence::kTernary, Associativity::RTL}},
+  };
+
+  std::unordered_map<TokenType, OperatorPrecedence> Precedence::kSpecialPrec {
```
dc03, line 109: This is meant for operators that don't really fit into any of the other categories, like array subscript being technically binary but with an extra token at the end, and function call being an n-ary operator where the opening parentheses is an "infix" operator.

PR #2307, opened 2022-06-26
emake: add -j flags to specify compile jobs

dc03:
add a `-j` flag to emake for that yoinky sploinky

### DIFF
--- a/CommandLine/emake/OptionsParser.cpp
+++ b/CommandLine/emake/OptionsParser.cpp
@@ -131,6 +131,7 @@ OptionsParser::OptionsParser() : _desc("Options")
     ("enigma-root", opt::value<std::string>()->default_value(fs::current_path().string()), "Path to ENIGMA's sources")
     ("codegen-only", opt::bool_switch()->default_value(false), "Only generate code and exit")
     ("run,r", opt::bool_switch()->default_value(false), "Automatically run the game after it is built")
+    ("jobs,j", opt::value<int>()->default_value(1), "The number of compile jobs to run simultaneously")
   ;
 
   _positional.add("input", 1);
@@ -245,6 +246,7 @@ std::string OptionsParser::APIyaml(const buffers::resources::Settings* currentCo
   std::string widgets = _rawArgs["widgets"].as<std::string>();
   std::string collision = _rawArgs["collision"].as<std::string>();
   std::string network = _rawArgs["network"].as<std::string>();
+  std::string jobs = std::to_string(_rawArgs["jobs"].as<int>());
   
   std::string eobjs_directory = fs::absolute(_rawArgs["workdir"].as<std::string>()).string();
   std::string codegen_directory = fs::absolute(_rawArgs["codegen"].as<std::string>()).string(); 
@@ -306,6 +308,7 @@ std::string OptionsParser::APIyaml(const buffers::resources::Settings* currentCo
   yaml += "extensions: " + _extensions + "\n";
   yaml += std::string("codegen-only: ") + (_rawArgs["codegen-only"].as<bool>() ? "true" : "false") + "\n";
   yaml += "enigma-root: " + _enigmaRoot + "\n";
+  yaml += "jobs: " + jobs + "\n";
 
   return yaml;
 }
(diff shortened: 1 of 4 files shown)


CompilerSource/settings-parse/parse_ide_settings.cpp
```diff
@@ -139,6 +140,9 @@ void parse_ide_settings(const char* eyaml)
   setting::automatic_semicolons   = settree.get("automatic-semicolons").toBool();
   setting::keyword_blacklist = settree.get("keyword-blacklist").toString();
 
+  // The number of compile jobs
+  num_make_jobs = settree.get("jobs").toString();
```
JoshDreamland, line 144: Probably worth round-tripping to int to make sure we don't pass anything weird to make.

CompilerSource/compiler/compile.cpp
```diff
@@ -725,6 +726,7 @@ int lang_CPP::compile(const GameData &game, const char* exe_filename, int mode)
   make += "NETWORKING=\""  + extensions::targetAPI.networkSys + "\" ";
   make += "PLATFORM=\"" + extensions::targetAPI.windowSys + "\" ";
   make += "TARGET-PLATFORM=\"" + compilerInfo.target_platform + "\" ";
+  make += "-j" + num_make_jobs + " ";
```
JoshDreamland, line 729: This will default the number of jobs to unbounded, which is likely to cause problems on some systems.

CompilerSource/settings-parse/parse_ide_settings.cpp
dc03: It is defined to be an int, the parsing should fail if it isn't
```cpp
("jobs,j", opt::value<int>()->default_value(1), "The number of compile jobs to run simultaneously")
```

CompilerSource/compiler/compile.cpp
dc03: Check previous comment, default value is 1. I tested without `-j` and it worked fine, using only 1 job

CommandLine/emake/OptionsParser.cpp
```diff
@@ -245,6 +246,7 @@ std::string OptionsParser::APIyaml(const buffers::resources::Settings* currentCo
   std::string widgets = _rawArgs["widgets"].as<std::string>();
   std::string collision = _rawArgs["collision"].as<std::string>();
   std::string network = _rawArgs["network"].as<std::string>();
+  std::string jobs = std::to_string(_rawArgs["jobs"].as<int>());
```
dc03, line 249: It is also parsed as an int before being used in the yaml @JoshDreamland

CompilerSource/compiler/compile.cpp
fundies: emake sends a default but lgm does not have one

CompilerSource/settings-parse/parse_ide_settings.cpp
JoshDreamland: This string isn't just built from emake flags; it can be read from elsewhere (e.g. when prepared in LGM), too (and indeed, LGM will not be specifying a value for it, so the YAML parser will come back with the empty string).

CompilerSource/settings-parse/parse_ide_settings.cpp
```diff
@@ -141,7 +142,16 @@ void parse_ide_settings(const char* eyaml)
...
-  num_make_jobs = settree.get("jobs").toString();
+  if (settree.exists("jobs")) {
+    const std::string &jobs = settree.get("jobs");
+    if (std::all_of(jobs.begin(), jobs.end(), [](char c) { return std::isdigit(c); })) {
+      num_make_jobs = settree.get("jobs").toString();
+    } else {
+      num_make_jobs = "1";
+    }
+  } else {
+    num_make_jobs = "1";
```
JoshDreamland, line 153: Might I suggest just setting `num_make_jobs = "1";` up front and allowing the main branch to override?

CompilerSource/compiler/compile.cpp
JoshDreamland: Personally I'd just default to empty and predicate this line (and 367) on `!num_make_jobs.empty()`. But I guess the default of "1" in `parse_ide_settings` is good enough.

JoshDreamland (approved):

PR #2313, opened 2022-07-21
Win32 TestHarness

WanderingNullPtr:
Added Win32 TestHarness.

Screenshot uses python Pillow library. Install using system python.


generate_report.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
JoshDreamland, line 1: This is a lot of CI files at the project root... I understand the .yml files need to be there for the CI services to pick up, but can we move the other utilities into a CI folder? (Probably under //CommandLine/ or //CommandLine/testing/)

reportstyle.css
```diff
@@ -0,0 +1,28 @@
+@import url('https://fonts.googleapis.com/css2?family=Titillium+Web&display=swap');
```
JoshDreamland, line 1: (This file in particular probably doesn't belong in the root of the repo...)

Compilers/Windows/TestHarness.ey
```diff
@@ -0,0 +1,25 @@
+%e-yaml
```
JoshDreamland, line 1: Custom compiler for the test harness? Neat idea.

JoshDreamland (commented):
Good work! I am concerned about what looks like copy/pasting from the existing test harness (this project already suffers from a lot of code duplication) and I'm worried that one of the changes reduced the scope of the testing, but otherwise, this looks to be in good shape! My comments are inline.

It would also be good to hammer this into a working state again before we pull it.

PR #2337, opened 2023-03-17
Fix GTest extension build fails on installing it by building from source on windows

### LINKED ISSUE #2338: GTest extension crashes on windows
GTest extension crashes when used inside a game ```cannot find -lgtest.dll: No such file or directory```

### Steps To Reproduce
1. Run a SOG file on windows ```$ ./emake.exe -r CommandLine/testing/SimpleTests/.sog -e "GTest" -o /tmp/test.exe
```

k0T0z:
Fix #2338 

GTest extension must be installed using ``pacman -S mingw-w64-x86_64-gtest`` to be working correctly with the current ``ENIGMAsystem/SHELL/Universal_System/Extensions/GTest/Makefile``, but if you installed GTest with another way (building it from source and copy-paste files), it won't work because you will need the ``.dll`` file for linking it dynamically which is not provided when you build gtest from source, you only got ``.a`` files.

Using pacman or building from source gives you ``.a`` files, which you can link statically to your exe and this PR provides that change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,10 @@
 /CompilerSource/stupidity-buffer/obj
 /ENIGMAsystem/SHELL/.vs/*
 /ENIGMAsystem/SHELL/.vscode/*
+/.vscode/*
+/CMakeFiles/*
+/ENIGMAsystem/SHELL/.build/*
+CMakeCache
 .DS_Store
 
 enigma-launch.bat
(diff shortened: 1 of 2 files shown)


RobertBColton (approved):
Looks good, glad they got back to me about this!
https://github.com/msys2/MINGW-packages/issues/5113

saifkandil:
cool, thanks robert

k0T0z:
I like the dynamic approach so let's close this as it's not needed anymore.

PR #2351, opened 2023-08-09
GSoC Final Evaluation: Implementing (de)serialization routines for generic data types

FaresAtef1:


ENIGMAsystem/SHELL/Universal_System/Resources/polygon.h
FaresAtef1, line 1: To eliminate a warning.

ENIGMAsystem/SHELL/Universal_System/Resources/polygon.h
```diff
@@ -1,5 +1,5 @@
 /** Copyright (C) 2021 Nabeel Danish
-/** 
+***
```
FaresAtef1, line 2: To eliminate a warning.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization.h
```diff
@@ -0,0 +1,817 @@
...
+template <typename T>
+inline T internal_deserialize_numeric(std::byte *iter) {
+  if constexpr (std::is_integral_v<T>) {
+    return internal_deserialize_integral<T>(iter);
+  } else if constexpr (std::is_floating_point_v<T>) {
+    return internal_deserialize_floating<T>(iter);
+  } else {
+    static_assert(always_false<T>, "'internal_deserialize_numeric' takes either integral or floating types");
+  }
+}
```
dc03, line 180: Going forward, I think it may be a good idea to either:
- have six separate files for the six functions where each file contains implementations for all types
- have (roughly) separate files for each type implementing all of the functions per file

Of these, I like the second option more because it groups all the functionality for either a type or a set of related types together without noise from implementations for other types. In this case, I think it'd be a good idea to have all the functions for the integral and floating types in one file.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization.h
```diff
@@ -0,0 +1,817 @@
...
+  if constexpr (has_internal_serialize_fn_free_function<std::decay_t<T>>) {
+    return internal_serialize_fn(value);
+  } else if constexpr (has_serialize_method_v<std::decay_t<T>>) {
+    return value.serialize();
+  } else {
+    static_assert(always_false<T>,
+                  "'serialize' takes 'variant', 'var', 'std::string', bool, integral, floating types, std::vector, "
+                  "std::map or std::complex");
+  }
+}
```
dc03, line 474: I think the most important thing you should do before adding functions for more types is to update all of these to use the const-reference parameters and pattern match instead of using `is_same_v` so that it is less effort to move them later.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization.h
```diff
@@ -0,0 +1,817 @@
...
+
+template <typename T>
+typename std::enable_if<std::is_pointer_v<std::decay_t<T>>, T>::type inline internal_deserialize_fn(std::byte *iter) {
+  return nullptr;
+}
+
+template <typename T>
+typename std::enable_if<std::is_same_v<std::string, std::decay_t<T>>, T>::type inline internal_deserialize_fn(
```
dc03, line 482: May be a good idea to create a trait for this pattern, by calling it `is_type_t` or even `is_t` to make the code much terser and easier to read.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization.h
```diff
@@ -0,0 +1,817 @@
...
+  if constexpr (has_internal_resize_buffer_for_fn_free_function<std::decay_t<T>>) {
+    internal_resize_buffer_for_fn(buffer, value);
+  } else if constexpr (has_byte_size_method_v<std::decay_t<T>>) {
+    internal_resize_buffer_using_byte_size(buffer, value);
+  } else {
+    internal_resize_buffer_for_value(buffer, value);
+  }
+}
+
+// A varint encoding version of the integer functions
```
dc03, line 672: I should've done this in my project itself, but we can probably drop the VBR encoding functions which are commented out because these will probably never be used.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization_fwd_decl.h
```diff
@@ -0,0 +1,252 @@
...
+template <typename T>
+struct complex_inner_type;
+
+template <typename T>
+struct complex_inner_type<std::complex<T>> {
+  using type = T;
+};
+
+template <typename T>
+using complex_inner_type_t = typename complex_inner_type<T>::type;
```
dc03, line 108: Hmm, my plan for this file was to have just six lines of code: the six forward declarations for the functions that the user can implement. This way it is automatically documenting and we can just point users here to tell them what functions to implement. Its not required to forward-declare the templates which are specialized for the various types, because the user will get them anyways when `serialization.h` is included. For the type traits, I would suggest making a `type_traits.h` header file within this directory itself.

ENIGMAsystem/SHELL/Universal_System/Resources/polygon.h
dc03: I think to match the comment style it is probably best to have ` **` instead of `***` (space instead of star)

ENIGMAsystem/SHELL/Universal_System/Serialization/detect_size.h
```diff
@@ -0,0 +1,82 @@
...
+template <typename T>
+constexpr static inline bool always_false = false;
+
+inline std::size_t byte_size(const variant &value) { return variant_size(value); }
+
+inline std::size_t byte_size(const var &value) { return var_size(value); }
+
+template <typename T>
+inline std::size_t byte_size(const std::vector<T> &value) {
+  return value.size() * sizeof(T);
```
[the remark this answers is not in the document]
dc03: Actually, given this condition, I think the function for `lua_table` is also wrong, because I wrote it only thinking about fixed-size types as well, so it would probably be a good idea to fix that as well.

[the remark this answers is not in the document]
FaresAtef1: ok, I will fix them.

ENIGMAsystem/SHELL/Universal_System/Serialization/detect_size.h
```diff
@@ -0,0 +1,82 @@
...
+}
+
+template <typename T>
+inline std::size_t byte_size(const std::complex<T> &value) {
+  return sizeof(T) * 2;
+}
+
+template <typename T>
+inline std::size_t byte_size(const lua_table<T> &value) {
+  return enigma_internal_sizeof_lua_table(value);
```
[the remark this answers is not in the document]
FaresAtef1: ok, and so I will do the same thing with `internal_resize_buffer_for_type` and `internal_resize_buffer_for_fn`.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization_fwd_decl.h
FaresAtef1: Cool solution, I will do that.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization.h
FaresAtef1: ok, I will do that.

ENIGMAsystem/SHELL/Universal_System/Serialization/serialization.h
FaresAtef1: Nice idea.

ENIGMAsystem/SHELL/Universal_System/Extensions/GTest/include.h
```diff
@@ -61,7 +61,7 @@ namespace enigma_user {
 template<typename T, typename U>
 void gtest_check_eq(T a, U b, std::string message, std::string aE, std::string bE, bool assert) {
   enigma::gtest_binary("(" + aE + ") == (" + bE + ")", toString(a), toString(b), aE, bE,
-                       "is equal to",                  a == b, assert, message, "EDL", 1);
+                       "is equal to",                  a == static_cast<T> (b), assert, message, "EDL", 1);
```
FaresAtef1, line 64: To eliminate a warning.

ENIGMAsystem/SHELL/Universal_System/Resources/polygon.h
dc03: > I think to match the comment style it is probably best to have ` **` instead of `***` (space instead of star)

Oh, nevermind this, its been a while since I looked at this code and the comment style is indeed `***`. Oops.

FaresAtef1: np.

PR #2363, opened 2023-11-12
Update Steamworks SDK from v1.57 to v1.58a

k0T0z:
The next PR after #2350 and #2343.

This PR contains the following changes:

 1. Update SDK from v1.57 to v1.58a.
 2. Improved the Shell Makefile (the part related to the Steamworks extension).
     - Moved building the fake API process inside a separate Makefile specific to the fake API.
 3. Cleaned some comments including flushing the Steamworks Makefile.
 4. Fixed the README.md file.
 5. Improved documentation.
 6. Created a separate Makefile for the fake Steamworks API.
 7. Added v1.58a official headers.
 8. Fixed .gitignore file.


ENIGMAsystem/SHELL/Universal_System/Extensions/Steamworks/gameclient/steambinder/steambinder.h
```diff
@@ -39,21 +40,34 @@
 #error "ENIGMA_STEAMWORKS_EXTENSION_ROOT is not defined."
 #endif  // ENIGMA_STEAMWORKS_EXTENSION_ROOT
 
+#ifndef STEAMWORKS_API_VERSION
+#error "STEAMWORKS_API_VERSION is not defined."
+#endif  // STEAMWORKS_API_VERSION
+
+#define STR(x) #x
+#define S_VERSIONED_ACCESSOR_NAME(name) STR(name)
```
k0T0z, line 48: Ig this should be ``VERSIONED_ACCESSOR_NAME``.

JoshDreamland (approved):
This is extremely chatty; are we going to incur another 3K lines every time Steam pushes an update? If so, we need to move this out of our code base. Used to be ENIGMA was small enough that deep clones were zippy; I'd like to keep it that way.

k0T0z:
Got it

k0T0z:
Thank you, what I intend to do now is to keep only 2 versions of Steamworks on the repository. This means on each new version, I will remove an existing one (older one).

JoshDreamland:
Removing files from a Git repository *increases* the size of the Git repository (except when doing shallow clones).

k0T0z:
I thought you were talking about the size of the code on the repository. We can't escape for having a big repository anyway, can't we?

JoshDreamland:
Some size inflation is of course required in any living project. What we want to try to avoid is every feature creating thousands of lines of traffic each year. Most people won't need Steam API integration, so it's better to keep the traffic needed for it out of the main repo. Part of the value-add of the Extensions system is that systems that have large dependencies don't inflict those dependencies on everyone who installs ENIGMA.

What ENIGMA really needs for dealing with this is proper package manager integration so that you can install the dependencies of extensions you care about automatically from the UI. But we're not there yet in terms of project maturity; we need to grow past Msys's native packages breaking monthly, first.

k0T0z:
Seems fine, but note that I already contacted MSYS before starting this project and they can not provide it. I think another solution to this is to create a separate repo for it and use it as a submodule inside the engine or even leave it separated from the engine and in the CI we can do ``git clone --depth 1`` to build it with the engine, what do you think?

JoshDreamland:
A submodule sounds perfectly fine. Where are you obtaining these sources from? There just a download link on Steam's site?

k0T0z:
Yeah, the link has a zip file that contains the headers as well as binaries.

PR #2372, opened 2023-12-13
#2371 fix:Steamworks logo is now showing

Daksh-10:
In reference to #2371 

On going through the code, I found out in the yaml file of steamworks, the icon tag didn't had the correct file name, I changed it, also the svg wasnot working, I clearly didn't know why so I rather did put a steamwork.png image, also I added a start.sh file, so ubuntu and debian users don't need to install it separately.

PR #2373, opened 2023-12-16
Update steamworks.svg

Raghvendra9402:
the logo is not showing because the styling of <svg>. previously the color is white(#fff) because it was blending with the background, i change the styling to the black color.


k0T0z:
@Raghvendra9402 Thank you for your contribution. This seems to be a duplicate of #2372.

PR #2380, opened 2024-03-01
Finish Matchers Work and Semicolon Modifications

FaresAtef1:
This PR finalizes the tasks initiated in #2375. It addresses the remaining semicolon issues and incorporates additional fixes to the parser.

### Modified Expressions
- Unary Prefix
- Binary

### Testing
- Add new if statement tests.
- Create a new type cast method `AST::Node::As`. 
- Create a new header file for all the matchers.

### To be done
- [x] Fix the infinite loop error in the binary expression.
- [x] Modify the matchers for `for loop`.
- [x] Use AST cast routine instead of `dynamic_cast`.
- [x] Handle the following cases:

```cpp
for(i=0;i<3;i++)

{k+;}

fn(i , j+2 ;)

fn(i , j ;)

```


CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -35,7 +35,244 @@ struct ParserTester {
...
   ASSERT_EQ(node->type, AST::NodeType::IDENTIFIER);
-  ASSERT_EQ(dynamic_cast<AST::IdentifierAccess *>(node)->name.content, name);
+  ASSERT_EQ(node->As<AST::IdentifierAccess>()->name.content, name);
+}
+
+std::string ExpectedMsg = "";
+
+MATCHER_P(IsDeclaration, decls, "") {
+  bool b1 = arg->type == AST::NodeType::DECLARATION;
+  if (!b1) {
```
FaresAtef1, line 45: For this and all remaining matchers, this check will be the first one before the casting to ensure that we do not encounter a `Segmentation fault`.

CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -35,7 +35,244 @@ struct ParserTester {
...
+    ExpectedMsg = "From IsDeclaration Matcher: NodeType = DECLARATION\n";
+    *result_listener << "got NodeType = " << AST::NodeToString(arg->type) << "\n";
+    return 0;
+  }
+
+  auto *decl = arg->template As<AST::DeclarationStatement>();
```
FaresAtef1, line 51: Now, there is no need to check whether this pointer is not equal to `nullptr`, as the previous condition is true.

CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -35,7 +35,244 @@ struct ParserTester {
 
 void assert_identifier_is(AST::Node *node, std::string_view name) {
   ASSERT_EQ(node->type, AST::NodeType::IDENTIFIER);
-  ASSERT_EQ(dynamic_cast<AST::IdentifierAccess *>(node)->name.content, name);
+  ASSERT_EQ(node->As<AST::IdentifierAccess>()->name.content, name);
+}
+
+std::string ExpectedMsg = "";
+
+MATCHER_P(IsDeclaration, decls, "") {
```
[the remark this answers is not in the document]
FaresAtef1: Good idea; it will also be a complete mess if we want to add new matchers (I plan to do this) in the same file.

JoshDreamland (approved):
This is looking really good. Thank you! I've picked a couple nits but I wouldn't stall the works on them.

CommandLine/emake-tests/Parsing/matchers.h
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  auto *expr = cast->expr->template As<AST::Node>();
+  if (!expr) {
+    ExpectedMsg += "From IsCast Matcher: expr isn't nullptr\n";
+    *result_listener << "got expr = nullptr\n";
+    return 0;
+  }
+
+  bool b1 = cast->ft.def == jdi::builtin_type__int, b2 = cast->ft.flags == 0, b3 = cast->ft.decl.components.size() == 0,
```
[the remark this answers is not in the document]
FaresAtef1: Yeah sure, I totally forgot about it when I did the same with `IsDeclaration` matcher.

CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -200,26 +200,25 @@ TEST(ParserTest, Declarator_1) {
...
-  EXPECT_EQ((first++)->type, jdi::ref_stack::RT_POINTERTO);
-  EXPECT_EQ((first++)->type, jdi::ref_stack::RT_POINTERTO);
-  EXPECT_EQ(test2.lexer.ReadToken().type, TT_ENDOFCODE);
+  ASSERT_EQ(ft2.decl.name.content, "y");
+  ASSERT_EQ((first++)->type, jdi::ref_stack::RT_POINTERTO);
+  ASSERT_EQ((first++)->type, jdi::ref_stack::RT_MEMBER_POINTER);
+  ASSERT_EQ((first++)->type, jdi::ref_stack::RT_ARRAYBOUND);
+  ASSERT_EQ((first++)->type, jdi::ref_stack::RT_POINTERTO);
+  ASSERT_EQ((first++)->type, jdi::ref_stack::RT_POINTERTO);
+  ASSERT_EQ(test2.lexer.ReadToken().type, TT_ENDOFCODE);
```
FaresAtef1, line 209: To avoid the segmentation fault.

CompilerSource/parsing/parser.cpp
```diff
@@ -2102,19 +2102,25 @@ std::unique_ptr<AST::Node> ParseCFStmtBody() {
...
   std::vector<std::unique_ptr<AST::Node>> statements{};
 
-  while (token.type != TT_ENDOFCODE) {
-    if (auto node = TryParseStatement()) {
-      statements.push_back(std::move(node));
-    } else if (token.type == TT_SEMICOLON) {
-      token = lexer->ReadToken();
-    } else break;
+  while (token.type != TT_ENDBRACE) {
+    if (token.type == TT_BEGINBRACE) {
```
JoshDreamland, line 2117: Can these calls be replaced with a ParseStatementOrBlock? I feel like we have one of those or, if not, we need them.

JoshDreamland (approved):

Issue #2365, opened 2023-11-15
Incorrect way of getting the target Steam Leaderboards


k0T0z:
Regarding the [leaderboard API](https://github.com/YoYoGames/GMEXT-Steamworks/wiki/leaderboards), when you call functions like ``steam_upload_score()``, the function starts to find the leaderboard at first as in line 233 https://github.com/enigma-dev/enigma-dev/blob/3918af3dc40734465b355a7d52cb9f9f93f5a8d6/ENIGMAsystem/SHELL/Universal_System/Extensions/Steamworks/leaderboards.cpp#L233.

This is incorrect as the first call for this function will be useless and will fail because at first there is no leaderboard handle stored (this check https://github.com/enigma-dev/enigma-dev/blob/3918af3dc40734465b355a7d52cb9f9f93f5a8d6/ENIGMAsystem/SHELL/Universal_System/Extensions/Steamworks/gameclient/gc_leaderboards.cpp#L124 will exit the function).

This is the case for all of these functions:
 - steam_upload_score_ext()
 - steam_download_scores()
 - steam_download_scores_around_user()
 - steam_download_friends_scores()

As ENIGMA is single-threaded, I would like to handle this in a way such that, those functions will succeed on their first call.

PR #2385, opened 2024-03-18
fix leaderboards func to address #2365

YoungHypo:
### Description
Create `std::list<LeaderboardRequest> LBRequestList` for all download or upload leaderboards requests, and a `LeaderboardRequest` structure object to store the relevant parameters for the calls. 
### Example:
Take the `steam_download_scores` function [leaderboards.cpp](https://github.com/enigma-dev/enigma-dev/blob/master/ENIGMAsystem/SHELL/Universal_System/Extensions/Steamworks/leaderboards.cpp) as an example, first add the leaderboard request of `download_scores` func to `LBRequestList`. When `find_leaderboard` func completes the find operation, the list will release the download request for execution. Therefore, the "download" of the leaderboard entry always follows the operation of finding the leaderboard, and the two remain synchronized.


YoungHypo:
@k0T0z

PR #2387, opened 2024-03-27
Add ability to set gRPC executable PATH

k0T0z:
When building gRPC from source and installing it using CMake (which is recommended by gRPC owners, see [grpc.io/blog/installation](https://grpc.io/blog/installation/#want-to-try-it)) the default path in ENIGMA to grpc_cpp_plugin is not applicable anymore, 

```
/usr/bin/grpc_cpp_plugin: program not found or is not executable
```

why not just add a way for passing it instead of hardcoding? This PR will do it.


k0T0z:
Close as merged in #2399

PR #2395, opened 2024-06-17
`ParseStatementOrBlock` and `TryParseDeclOrTypeExpression`

FaresAtef1:
In this PR, I added new tests and 2 new functions `ParseStatementOrBlock` and `TryParseDeclOrTypeExpression`.

### DIFF
--- a/CommandLine/emake-tests/Parsing/parser-test.cpp
+++ b/CommandLine/emake-tests/Parsing/parser-test.cpp
@@ -40,6 +40,56 @@ void assert_identifier_is(AST::Node *node, std::string_view name) {
 }
 
 TEST(ParserTest, Basics) {
+  ParserTester test{"(x ? y : z ? a : (z[5](6)));"};
+
+  auto node = test->TryParseStatement();
+  ASSERT_EQ(node->type, AST::NodeType::PARENTHETICAL);
+
+  auto *expr = node->As<AST::Parenthetical>()->expression.get();
+  ASSERT_EQ(expr->type, AST::NodeType::TERNARY_EXPRESSION);
+
+  auto *ternary = expr->As<AST::TernaryExpression>();
+  auto *cond = ternary->condition.get();
+  auto *true_ = ternary->true_expression.get();
+  auto *false_ = ternary->false_expression.get();
+  assert_identifier_is(cond, "x");
+  assert_identifier_is(true_, "y");
+
+  ASSERT_EQ(false_->type, AST::NodeType::TERNARY_EXPRESSION);
+
+  ternary = false_->As<AST::TernaryExpression>();
+
+  cond = ternary->condition.get();
+  true_ = ternary->true_expression.get();
+  false_ = ternary->false_expression.get();
+
+  assert_identifier_is(cond, "z");
+  assert_identifier_is(true_, "a");
+
+  ASSERT_EQ(false_->type, AST::NodeType::PARENTHETICAL);
+  expr = false_->As<AST::Parenthetical>()->expression.get();
+  ASSERT_EQ(expr->type, AST::NodeType::FUNCTION_CALL);
+  auto *function = expr->As<AST::FunctionCallExpression>();
+  auto *called = function->function.get();
+  auto *args = &function->arguments;
+
+  ASSERT_EQ(called->type, AST::NodeType::BINARY_EXPRESSION);
+
+  auto *bin = called->As<AST::BinaryExpression>();
+  ASSERT_EQ(bin->operation, TokenType::TT_BEGINBRACKET);
+  assert_identifier_is(bin->left.get(), "z");
+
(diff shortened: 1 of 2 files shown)


CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -40,6 +40,56 @@ void assert_identifier_is(AST::Node *node, std::string_view name) {
...
+  auto *false_ = ternary->false_expression.get();
+  assert_identifier_is(cond, "x");
+  assert_identifier_is(true_, "y");
+
+  ASSERT_EQ(false_->type, AST::NodeType::TERNARY_EXPRESSION);
+
+  ternary = false_->As<AST::TernaryExpression>();
+
+  cond = ternary->condition.get();
+  true_ = ternary->true_expression.get();
```
JoshDreamland, line 63: nit: Usually variables ending in `_` indicate private fields, so you might want to prefer a name like `true_exp`. 

As another odd point to note: names containing `__` (two underscores) or starting in an underscore *and* capital letter are reserved for the compiler/standard libraries.

CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -884,20 +1039,35 @@ TEST(ParserTest, SwitchStatement_5_NoSemicolon) {
...
-  auto *block = dynamic_cast<AST::CodeBlock *>(node.get());
+  auto *block = node->As<AST::CodeBlock>();
   ASSERT_EQ(block->statements.size(), 4);
   ASSERT_EQ(block->statements[0]->type, AST::NodeType::DECLARATION);
   ASSERT_EQ(block->statements[1]->type, AST::NodeType::DECLARATION);
   ASSERT_EQ(block->statements[2]->type, AST::NodeType::DECLARATION);
   ASSERT_EQ(block->statements[3]->type, AST::NodeType::FUNCTION_CALL);
 }
 
+TEST(ParserTest, CodeBlock_2) {
```
JoshDreamland, line 1056: Thanks for adding!

CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -315,6 +445,31 @@ TEST(ParserTest, Declarations) {
...
+
+  auto node = test->TryParseStatement();
+  EXPECT_EQ(test->current_token().type, TT_ENDOFCODE);
+  EXPECT_EQ(test.lexer.ReadToken().type, TT_ENDOFCODE);
+
+  ASSERT_EQ(node->type, AST::NodeType::DECLARATION);
+  auto *decls = reinterpret_cast<AST::DeclarationStatement *>(node.get());
+  EXPECT_EQ(decls->def->flags & jdi::DEF_TYPENAME, jdi::DEF_TYPENAME);
+
+  EXPECT_EQ(decls->declarations.size(), 3);
```
JoshDreamland, line 459: FYI: Matchers can actually assert that sequences contain a series of other matchers, so if we had one that matched on FullType, you could use
```cpp
ElementsAre(
    FullTypeWith(NoInitiailzer(), DeclarationWithFlags(jdi::DEF_TYPENAME), 
                 RefStackWith(DeclaratorNode::Kind::POINTER_TO)),
    FullTypeWith...)
```

You can get pretty creative with the Matchers interface. Consider it optional, though as you write more of these tests, I think it'll come in handy.

CompilerSource/parsing/parser.cpp
```diff
@@ -1950,7 +1950,29 @@ std::unique_ptr<AST::Node> TryParseControlExpression(SyntaxMode mode_) {
   }
 }
 
+std::unique_ptr<AST::Node> MaybeDeclaration() {
```
JoshDreamland, line 1953: I worry this doesn't give us the ability to indicate that parsing went wrong on our return. It seems that all branches either do not consume tokens or end in another TryParse, so I think this is okay, except if `TryParseEitherFunctionalCastOrDeclaration` were to return `nullptr` without consuming anything, we'd have accidentally swallowed the `::` token this routine was called with. Does that make sense? If `TryParseEitherFunctionalCastOrDeclaration` will necessarily return non-null when the token is `::`, we're fine.

In that case, I think this should be renamed to TryParseSomething or another, but then, it seems that what we handle here is just functional casts, declarations, and global new/delete, which might just be worth adding to `TryParseEitherFunctionalCastOrDeclaration` rather than creating a new method. That function's name is already Jupiterian, so I might recommend just `TryParseDeclarationOrTypeExpression`.

What do you think? I'm sure Dhruv has stronger opinions than either of us. :slightly_smiling_face:

FaresAtef1: Agreed, didn't think of this case.

JoshDreamland:
Thanks a bunch! Let's get this merged...

PR #2396, opened 2024-06-18
GSoC Midterm Evaluation: Initiate the Pretty Printer

### LINKED ISSUE #2325: Implement the new parser
# Finish creation of the new EDL parser

Last year's project, titled "Data Buffers/Serialization", led to a complete
rewrite of the EDL (ENIGMA Development Language) parser. As it stands, the
parser can parse most of the EDL language, with some features which are
incomplete marked as "Unimplemented".

- The project would begin by finishing up the unimplemented features, and adding
  more test cases to the [parser tests](https://github.com/enigma-dev/enigma-dev/blob/AST-Generation/CommandLine/emake-tests/Parsing/parser-test.cpp),
  to try and achieve maximal coverage.
- The issues mentioned in [issue #2316](https://github.com/enigma-dev/enigma-dev/issues/2316)
  would be fixed possibly by implementing a pass which walks the AST to modify
  it as per the issues mentioned.
- The parser would then be implemented into ENIGMA itself, by replacing calls
  to the old parser with calls to the new one. This would also involve writing
  routines to lower the AST into ENIGMA's data structures.

Mentors: Josh, Greg
Difficulty: Medium/Hard
Expected size: 250h - 350h
Skills required: C++ fundamentals, compiler fundamentals

FaresAtef1:
## Midterm GSoC Update

Closes #2325.
This pull request contains updates for my Google Summer of Code (GSoC) project at the midterm evaluation stage.

## Finished Functions
- [x] `VisitCodeBlock`
- [x] `VisitBinaryExpression`
- [x] `VisitFunctionCallExpression`
- [x] `VisitUnaryPrefixExpression`
- [x] `VisitUnaryPostfixExpression`
- [x] `VisitTernaryExpression`
- [x] `VisitFullType`
- [x] `VisitSizeofExpression`
- [x] `VisitAlignofExpression`
- [x] `VisitCastExpression`
- [x] `VisitParenthetical`
- [x] `VisitArray`
- [x] `VisitIdentifierAccess`
- [x] `VisitLiteral`
- [x] `VisitIfStatement`
- [x] `VisitForLoop`
- [x] `VisitWhileLoop`
- [x] `VisitDoLoop`
- [x] `VisitCaseStatement`
- [x] `VisitDefaultStatement`
- [x] `VisitSwitchStatement`
- [x] `VisitReturnStatement`
- [x] `VisitBreakStatement`
- [x] `VisitContinueStatement`
- [ ] `VisitWithStatement`
- [x] `VisitBraceOrParenInitializer`
- [x] `VisitAssignmentInitializer`
- [x] `VisitInitializer`
- [x] `VisitNewExpression`
- [x] `VisitDeleteExpression`
- [x] `VisitDeclarationStatement`

## Testing
- [x] Create a testing method for the pretty printer

## Organization
- [x] Organize if-else ladders in `pretty-printer.cpp`
- [x] Organize `parser.h`

## To Be Done Before Merging
- [x] Return booleans from all the visitor functions
- [x] Solve the [18446744073709551615] error
- [x] Solve types flags error
- [x] Store the operation for `BinaryExpression` 
- [x] Print escape characters
- [x] `signed int` Duplicate usage of flags

## What's Next
GSoC Final Evaluation:
- [ ] Parse `"'\''"` and `"'\"'"`
- [ ] `VisitWithStatement`
- [ ] lambdas
- [ ] "local" and "global"


CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -0,0 +1,1343 @@
...
+        jdi::builtin_flag__short->mask,    jdi::builtin_flag__long->mask,         jdi::builtin_flag__long_long->mask,
+        jdi::builtin_flag__restrict->mask, jdi::builtin_typeflag__override->mask, jdi::builtin_typeflag__final->mask};
+
+    std::vector<std::string> flags_names = {"const",  "static",    "volatile", "mutable",  "register",
+                                            "inline", "complex",   "unsigned", "signed",   "short",
+                                            "long",   "long long", "restrict", "override", "final"};
+
+    for (std::size_t i = 0; i < flags_values.size(); i++) {
+      if ((ft.flags & flags_masks[i]) == flags_values[i]) {
+        if (i != 8 || (i == 8 && ft.def->name == "char")) {
```
FaresAtef1, line 387: In most cases, `signed` is redundant, but it is most commonly used with `char`.

CompilerSource/parsing/parser.cpp
```diff
@@ -1723,6 +1692,7 @@ std::unique_ptr<AST::Node> TryParseOperand() {
         TryParseTypeSpecifier(&type);
         if (token.type == TT_BEGINPARENTH) {
           auto tok = token;
+          token = lexer->ReadToken();
```
FaresAtef1, line 1695: We need to consume the token here because if we don't, `TryParseExpression` will treat it as a `PARENTHETICAL`,
 consume the `)` token, and the subsequent `require_token` will fail.

CompilerSource/parsing/parser.cpp
```diff
@@ -383,15 +383,12 @@ bool AstBuilder::next_maybe_functional_cast() {
 }
 
 std::unique_ptr<AST::DeclarationStatement> AstBuilder::parse_declarations(
-    AST::DeclarationStatement::StorageClass sc, jdi::definition *def,
-    AST::DeclaratorType decl_type,
-  bool parse_unbounded, std::vector<AST::DeclarationStatement::Declaration> decls, bool already_parsed_first ) {
+    AST::DeclarationStatement::StorageClass sc, FullType &ft, AST::DeclaratorType decl_type, bool parse_unbounded,
```
FaresAtef1, line 386: I modified this function to accept the entire `FullType`, not just the `definition`, because we also need the flags. The original function assigned the same default flag (`signed`) to all declarations, as it didn't account for the parsed flags.

CompilerSource/parsing/parser.cpp
```diff
@@ -990,7 +987,11 @@ void AstBuilder::TryParseTypeSpecifier(FullType *type) {
         if (contains_decflag_bitmask(type->flags, "unsigned") && token.type == TT_SIGNED) {
           herr->Error(token) << "Conflicting use of 'unsigned' and 'signed' in the same type specifier";
         } else if (contains_decflag_bitmask(type->flags, token.content)) {
-          herr->Warning(token) << "Duplicate usage of flags in type specifier";
+          if (first_signed && token.type == TT_SIGNED) {
```
FaresAtef1, line 990: The `first_signed` bool is used to accept declarations like `signed int x`, because the `signed flag` is the default. Without this flag, encountering the `signed` token would be considered a duplicate usage of flags.

CompilerSource/parsing/parser.cpp
```diff
@@ -1494,11 +1494,8 @@ std::unique_ptr<AST::Node> AstBuilder::TryParseOperand() {
   switch (token.type) {
     case TT_BEGINBRACE: case TT_ENDBRACE:
     case TT_ENDPARENTH: case TT_ENDBRACKET:
-    case TT_ENDOFCODE:
+    case TT_ENDOFCODE: case TT_SEMICOLON:
       return nullptr;
-
-    case TT_SEMICOLON:
-      return std::make_unique<AST::CodeBlock>();
     case TT_COLON:
```
FaresAtef1, line 1499: I need to know if it found `;` so I can handle cases like `x++++;`. This can be determined by the `nullptr`.

CompilerSource/parsing/parser.cpp
JoshDreamland: No need to handle that here. It's not *syntactically* invalid to say `signed signed long unsigned signed int;`—the parser can *still build a tree.* We need to check that in a separate semantic checking layer afterward. Otherwise it creates a lot of work and awkward code flow here.

Another thing to check for in this layer: Does the "local" or "global" keyword appear in a type cast?

CompilerSource/parsing/parser.cpp
JoshDreamland: Either is fine. It normally makes sense to let TryParseExpression handle it, except we've already decided that the expression is a functional cast here, so you're right to avoid letting it consume this token.

I suspect that this code is actually still wrong, though, as it seems to treat identifiers this same way. I think it's getting away with it right now because we don't have any tests for lambdas, which would mean that identifiers here should be treated as function calls.

FaresAtef1: Great, I will take this into consideration. I will pay more attention to lambdas in the upcoming PR.

CompilerSource/parsing/parser.cpp
FaresAtef1: Okay, I removed them. Regarding local and global, I found that they need some extra work beyond just the type cast check, so I think it would be better to address them in the upcoming PR.

CompilerSource/parsing/parser.h
```diff
@@ -2,14 +2,263 @@
...
 #include "lexer.h"
+#include "precedence.h"
+#include "settings.h"
+#include "tokens.h"
+
+#include <JDI/src/System/builtins.h>
+#include <memory>
 
 namespace enigma::parsing {
+class AstBuilder {
```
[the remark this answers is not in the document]
FaresAtef1: Done.

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -0,0 +1,733 @@
...
+using namespace enigma::parsing;
+
+bool AST::Visitor::VisitIdentifierAccess(IdentifierAccess &node) {
+  print(std::string(node.name.content));
+  return true;
+}
+
+bool AST::Visitor::VisitLiteral(Literal &node) {
+  std::string value = std::get<std::string>(node.value.value);
+  if (value == "'\n'") {
```
JoshDreamland, line 30: This won't pass much more than your test cases.

EDL does some backflips to reflect how GML does things. We should have a setting somewhere for which language (C++ or GML) literals are inherited from. In this case, I don't think the onus should fall on the pretty printer to *decode* the string, but to encode it. So  basically, literals should know what type they have (char, string, int, float, etc) and what the final value is. So during *parse*, we should be checking if the input string was `"` or `'` (single or double-quote), and if single ***and*** literals are inherited from C++, interpret as a char literal (and warn if >1 char in the end). Similarly, that parse method should handle interpreting escapes. `\n` doesn't mean newline in GML; `#` means newline. This is dumb IMO, so it's configurable in ENIGMA.

So the parser (particularly the lexer) is already checking those settings (where literals are inherited from, where escapes are inherited from). It already uses those settings to determine whether `'a'` is a string literal (`TT_STRINGLIT`) or char literal (`TT_CHARLIT`). What I don't think it's doing is computing the *raw* value of these literals in the parser itself. In those routines, it should handle decoding those values:
```cpp
if (value.length() < 2 || value[0] != value[value.length() - 1]) return error(Internal error: Invalid string literal);
std::string str_value;
str_value.reserve(value.length() - 2);
if (options.use_escapes) {
  for (size_t i = 1; i < value.length() - 1; ++i) {
    if (value[i] == '\\') {
      if (++i >= value.length())
        return error(Internal error: lexer stopped parsing string literal early);
      switch (value[++i]) {
        case '\\': str_value += '\\'; break;
        case 'n': str_value += '\n'; break;
        // ...
        default: return error(Unkown escape); // or warn instead and insert it raw; do what your heart tells you
      }
    } else {
      str_value += value[i];
    }
  }
} else {
  for (size_t i = 1; i < value.length() - 1; ++i) {
    str_value += value[i] == '#' ? '\n' : value[i];
  }
}
```

Then in this routine, you encode the raw string:
```cpp
print(type == TT_CHARLIT ? '\'' : '\"');
for (char c : str_value) {
  if (c == '\\') {
    print("\\\\");
  } else if (c >= ' ' and c <= '~') {
    print(c);
  } else if (c == '\n') {
    print("\\n");
  } else if (c == '\r') {
    print("\\r");
  } else {
    std::ostringstream oss;
    oss << '\\' << std::oct << static_cast<int>(c);
    print(oss.str());
  }
}
print(type == TT_CHARLIT ? '\'' : '\"');
```

(Code is approximate/pseudo-code.)

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -0,0 +1,733 @@
...
+    print(value);
+  }
+  return true;
+}
+
+bool AST::Visitor::VisitParenthetical(Parenthetical &node) {
+  bool printed = false;
+  print("(");
+
+  printed = Visit(node.expression);
```
JoshDreamland, line 62: nit: Since you only really assign this once in most methods, it's probably better to just inline it (`if (!Visit(subnode)) return false;`).

FaresAtef1: Nice idea, it is cleaner now. I also introduced a `VISIT_AND_CHECK` macro.

CompilerSource/parsing/pretty_printer.cpp
FaresAtef1: Thanks, Josh, for this detailed explanation. I will handle this in the upcoming PR, along with the previously mentioned item in the to-do list: parse single and double quotes.

CompilerSource/parsing/pretty_printer.cpp
JoshDreamland: Heh, I don't think we need to include engineers' paranoia in this.  However, if you want to, please use the constant name rather than hard-coding `8` in the source.

The general rule of thumb is that no numbers should appear in the code outside of a constant definition except 0 and 1 (and even then, if there's a constant that more gracefully captures why you're using that number, you should prefer it...).

CompilerSource/parsing/parser.cpp
```diff
@@ -1922,32 +1892,22 @@ std::unique_ptr<AST::FunctionCallExpression> TryParseFunctionCallExpression(int
...
-        token = lexer->ReadToken(); // Consume the token
-        auto right = TryParseControlExpression(mode);
-        operand = std::make_unique<AST::BinaryExpression>(std::move(operand), std::move(right), oper.type);
-      }
-      // TODO: handle [] for array access and () for direct func call
-      return operand;
-    }
-    case SyntaxMode::GML:
-      return TryParseExpression(Precedence::kAll, nullptr);
+  if ((int)mode_ > 2) {
```
JoshDreamland, line 1895: Like below, this is another good place to use constants `(mode_ != STRICT && mode != QUIRKS)`. Otherwise, this looks like a very positive change. :slightly_smiling_face:

CompilerSource/parsing/parser.cpp
JoshDreamland: Possibly dangerous, or possibly obviously correct depending on how TryParseOperand is being used. Either way, good change. Be aware:

```cc
if (condition); else { do_something(); }
```

This is valid in C++. The `;` ends the `if`'s true branch, meaning `else` can happen immediately. I don't really care about this behavior in ENIGMA, but failing to accept it could cause misbehavior. From the name of this function, I don't think this case broke.

JoshDreamland (approved):
Good work! This change has come a long way. I've left some more comments inline; I think they're non-events and you can address the two of import whenever. Just be careful not to use constants' values everywhere, because your code will break if those constants change value. You want your code to fail to build in those cases, not break.

CompilerSource/parsing/pretty_printer.cpp
FaresAtef1: The purpose of this check is for testing. Imagine that I would need to add `signed` everywhere I used a type.

CompilerSource/parsing/parser.cpp
FaresAtef1: Nice case, but I think it needs to be handled in `TryParseStatement`.

CompilerSource/parsing/parser.cpp
FaresAtef1: Actually, there is another mode (`GML`); this is why I did that instead of a long condition.

PR #2398, opened 2024-07-22
GSoC Final Evaluation: Finish the Pretty Printer and Some Parser Work

FaresAtef1:


CompilerSource/parsing/ast.h
```diff
@@ -509,11 +521,15 @@ class AST {
 
   class Visitor {
     std::ofstream of;
+    bool print_type;
```
FaresAtef1, line 524: This will be used with `VisitLambdaExpression`. I prefer not to use it with `VisitFullType` because the default behavior for `VisitFullType` is to print the type, whereas for lambdas, the default is not to print the type.

CompilerSource/parser/collect_variables.cpp
```diff
@@ -68,6 +68,7 @@ class DeclGatheringVisitor : public AST::Visitor {
   const LanguageFrontend *const lang;
   ParsedScope *const parsed_scope;
   const NameSet &script_names;
+  CompileState *cs;
```
FaresAtef1, line 71: To be able to use `timeline_lookup`.

CompilerSource/parsing/lexer.cpp
```diff
@@ -396,8 +396,8 @@ Token Lexer::ReadRawToken() {
           herr->Error(Mark(spos, 1)) << "GML-style hex literal is trunucated";
           return ReadRawToken();
         }
-        while (!is_nybble(code[++pos]));
-        return Token(TT_HEXLITERAL, Mark(pos, pos - spos));
+        while (is_nybble(code[++pos]));
+        return Token(TT_HEXLITERAL, Mark(spos + 1, pos - spos - 1));
```
FaresAtef1, line 400: We don't need the `$`.

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -63,13 +69,31 @@ std::string AST::CppPrettyPrinter::GetPrintedCode() {
...
-  print(std::string(node.name.content));
+  std::string name = node.name.content;
+  if (is_script && name != "self") {
+    if (language_fe->is_shared_local(name)) {
+      print("enigma::glaccess(int(self))->" + name);
+    } else if (language_fe->global_exists(name)) {
+      print(name);
+    } else if (std::holds_alternative<jdi::definition *>(node.type) && std::get<jdi::definition *>(node.type)) {
+      print(name);
+    } else if (name.substr(0, 8) == "argument") {
```
FaresAtef1, line 80: Inside scripts, `argument0:15` are parameter names.

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -546,14 +572,18 @@ bool AST::CppPrettyPrinter::VisitNewExpression(AST::NewExpression &node) {
...
+  bool is_local = node.storage_class == DeclarationStatement::StorageClass::LOCAL;
+  if (is_global || is_local) {
     bool printed = false;
     for (std::size_t i = 0; i < node.declarations.size(); i++) {
       if (node.declarations[i].init) {
         if (printed) print(", ");
         std::string name = node.declarations[i].declarator->decl.name.content;
-        print(name + " = ");
+        if (is_global)
+          print("enigma::varaccess_" + name + "(int(global)) =");
```
FaresAtef1, line 584: This is useful in the case of `global int foo =12`.

ENIGMAsystem/SHELL/Universal_System/estring.cpp
```diff
@@ -172,6 +172,7 @@ double real(variant str) { return str.type ? atof(((string)str).c_str()) : (doub
 string ansi_char(char byte) { return string(1,byte); }
 string chr(char val) { return string(1,val); }
 int ord(string str)  { return str[0]; }
+int ord(char c) { return c; }
```
FaresAtef1, line 175: In the game [Traffic Simulator](https://enigma-dev.org/edc/games.php?game=94), there was a script that called `ord` with characters, so I thought it would be better to have this overload.

CompilerSource/parser/collect_variables.cpp
```diff
@@ -236,7 +236,8 @@ class DeclGatheringVisitor : public AST::Visitor {
       dectrip dtrip(type, prefix, suffix);
       if (is_global) parsed_scope->globals[name] = dtrip;
       if (is_local) parsed_scope->locals[name] = dtrip;
-      parsed_scope->declarations[name] = dtrip;
+      cs->add_dot_accessed_local(name);
```
FaresAtef1, line 239: It is useful for scripts, as it allows us to have `var &varaccess_foo(int x);` for all variables, enabling access to variables from within scripts using `enigma::varaccess_foo(int(self))`.

CompilerSource/parser/collect_variables.cpp
```diff
@@ -108,6 +108,7 @@ class DeclGatheringVisitor : public AST::Visitor {
       if (left->name.content == "global" && node.operation.type == enigma::parsing::TokenType::TT_DOT) {
         auto right = node.right->As<AST::IdentifierAccess>();
         parsed_scope->globals[right->name.content] = dectrip("var");
+        parsed_scope->locals[right->name.content] = dectrip("var");
```
FaresAtef1, line 196: This way, we can have both versions for any declared variable: a local one accessed through `foo` directly, and a global one accessed through `global.foo`.

CompilerSource/parser/collect_variables.cpp
```diff
@@ -234,6 +235,8 @@ class DeclGatheringVisitor : public AST::Visitor {
         prefix = ftype.substr(0, pos);
         suffix = ftype.substr(pos + name.length());
       }
+      prefix = std::regex_replace(prefix, std::regex("^ +| +$|( ) +"), "$1");
+      suffix = std::regex_replace(suffix, std::regex("^ +| +$|( ) +"), "$1");
```
FaresAtef1, line 239: Not doing that will prevent generating `case obj_0: return ((OBJ_obj_0*)inst)->val;` (for `local var val = 12;`) for some reason.

CompilerSource/parsing/parser.cpp
```diff
@@ -1856,7 +1918,7 @@ std::unique_ptr<AST::BinaryExpression> TryParseSubscriptExpression(int precedenc
     Token oper = token;
     token = lexer->ReadToken(); // Consume the operator
 
-    auto right = TryParseExpression(Precedence::kMin);
+    auto right = TryParseExpression(Precedence::kAll);
```
FaresAtef1, line 1921: To support multidimensional array access, like `game_field[a, b] = -1;`.

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -324,6 +340,9 @@ bool AST::CppPrettyPrinter::VisitReturnStatement(AST::ReturnStatement &node) {
   if (node.expression) {
     VISIT_AND_CHECK(node.expression);
   }
+  if (node.is_exit) {
+    print("0");
+  }
```
FaresAtef1, line 345: Our scripts return `variant`, so if `exit` is translated to `return;`, it will generate an error. Instead, it should be `return 0;`, as it is in the master branch.

CompilerSource/parsing/ast.h
```diff
@@ -508,80 +524,92 @@ class AST {
...
+    virtual bool VisitInitializer(Initializer &node){ return DefaultVisit(node); }
+    virtual bool VisitNewExpression(NewExpression &node){ return DefaultVisit(node); }
+    virtual bool VisitDeleteExpression(DeleteExpression &node){ return DefaultVisit(node); }
+    virtual bool VisitDeclarationStatement(DeclarationStatement &node){ return DefaultVisit(node); }
     virtual bool Visit(PNode &node) {
       return node->accept(*this);
     }
+  };
 
-    virtual ~Visitor() { of.close(); }
```
FaresAtef1, line 584: If I did that, some files used during the compilation process would be closed and not reopened, which would lead to errors.

Compilers/Linux/gcc.ey
```diff
@@ -12,7 +12,7 @@ Make-Vars:
   cc: gcc
   cxx: g++
   cppflags:
-  cxxflags: -fdiagnostics-color=always
+  cxxflags: -fdiagnostics-color=never
```
FaresAtef1, line 15: I will revert it to `always` before merging.

ENIGMAsystem/SHELL/Universal_System/Instances/with.h
```diff
@@ -1,39 +1,39 @@
...
+// **  applications created by its users, or damages caused by the environment     **
+// **  or programs made in the environment.                                        **
+// **                                                                              **
+// \********************************************************************************/
 
-#define with(x) \
-  for (enigma::iterator::with with(enigma::fetch_inst_iter_by_int(x)); \
-      enigma::instance_event_iterator; \
-      enigma::instance_event_iterator = enigma::instance_event_iterator->next)
+// #define with(x) \
```
FaresAtef1, line 29: It is now defined [here](https://github.com/enigma-dev/enigma-dev/blob/dad033e022ea334f85b569919eba334c123c11c3/CompilerSource/compiler/components/write_object_data.cpp#L858-L861).

CompilerSource/parsing/parser.cpp
```diff
@@ -2385,9 +2486,255 @@ std::unique_ptr<AST::WithStatement> ParseWithStatement() {
...
+      if (std::holds_alternative<jdi::definition *>(func->type)) {
+        def = std::get<jdi::definition *>(func->type);
+      } else {
+        node.RecursiveSubVisit(*this);
+        return false;
+      }
+      unsigned int min = 0;
+      unsigned int max = 0;
+      frontend->definition_parameter_bounds(def, min, max);
+      Token tok;
```
FaresAtef1, line 2508: There is a small problem: we don't have the line where it appears.

CommandLine/emake-tests/Parsing/parser-test-classes.h
```diff
@@ -30,10 +33,60 @@ struct ParserTester {
...
+  explicit ParserTester(std::string code, bool use_cpp)
       : context(&ParseContext::ForTesting(use_cpp)), lexer(std::move(code), context, &herr) {
     builder->initialize(&lexer, &herr);
   }
+
+  explicit ParserTester(std::string code) : context(&SetUp()), lexer(std::move(code), context, &herr) {
+    builder->initialize(&lexer, &herr);
+  }
+
+  static ParserTester CreateWithCpp(std::string code) { return ParserTester(std::move(code), true); }
```
RobertBColton, line 45: don't most compilers move automatically?

CommandLine/emake-tests/Parsing/parser-test-classes.h
```diff
@@ -30,10 +33,60 @@ struct ParserTester {
...
+      "inherit-increment-from: 0\n"
+      " \n"
+      "target-audio: OpenAL\n"
+      "target-windowing: xlib\n"
+      "target-compiler: gcc\n"
+      "target-graphics: OpenGL\n"
+      "target-widget: None\n"
+      "target-collision: BBox\n"
+      "target-networking: None\n"
+      ).c_str());
```
RobertBColton, line 76: I feel like this should be a struct rather than embedding strings.

CommandLine/emake-tests/Parsing/parser-test.cpp
```diff
@@ -114,7 +114,7 @@ TEST(ParserTest, Basics_NoSemicolon) {
 }
 
 TEST(ParserTest, SizeofExpression) {
-  ParserTester test{"sizeof 5"};  // we need to support sizeof (5)
+  ParserTester test = ParserTester::CreateWithCpp("sizeof 5");  // we need to support sizeof (5)
```
RobertBColton, line 117: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -23,34 +23,135 @@ using namespace enigma::parsing;
...
+      to_print += "\\f";
+    } else if (c == '\a') {
+      to_print += "\\a";
+    } else if (c == '\?') {
+      to_print += "\\?";
+    } else {
+      std::ostringstream oss;
+      oss << '\\' << std::oct << static_cast<int>(c);
+      to_print += oss.str();
+    }
```
RobertBColton, line 131: Feel like this could be cleaned up into a dict.

CompilerSource/parsing/lexer.cpp
```diff
@@ -318,14 +320,67 @@ TokenType Lexer::LookUpOperator(std::string_view op) {
...
+            break;
+          case 'a':
+            str_value += '\a';
+            break;
+          case '?':
+            str_value += '\?';
+            break;
+          default: 
+            herr->Error(Mark(spos, 1)) << "Unkown escape";
+            return lit;
```
RobertBColton, line 363: Same here, also feel this could be a map dict.

CompilerSource/parser/collect_variables.cpp
```diff
@@ -31,32 +32,119 @@
...
+  std::vector<std::size_t> flags_masks = {
+      jdi::builtin_flag__const->mask,    jdi::builtin_flag__static->mask,       jdi::builtin_flag__volatile->mask,
+      jdi::builtin_flag__mutable->mask,  jdi::builtin_flag__register->mask,     jdi::builtin_flag__inline->mask,
+      jdi::builtin_flag__Complex->mask,  jdi::builtin_flag__unsigned->mask,     jdi::builtin_flag__signed->mask,
+      jdi::builtin_flag__short->mask,    jdi::builtin_flag__long->mask,         jdi::builtin_flag__long_long->mask,
+      jdi::builtin_flag__restrict->mask, jdi::builtin_typeflag__override->mask, jdi::builtin_typeflag__final->mask};
+
+  std::vector<std::string> flags_names = {"const",  "static",    "volatile", "mutable",  "register",
+                                          "inline", "complex",   "unsigned", "signed",   "short",
+                                          "long",   "long long", "restrict", "override", "final"};
```
RobertBColton, line 82: what is the purpose of having these separate instead of an dict map or associative array?

I would at least assert that they are the same sizes and such.

CompilerSource/parsing/pretty_printer.cpp
```diff
@@ -63,13 +69,31 @@ std::string AST::CppPrettyPrinter::GetPrintedCode() {
...
+  }
   return true;
 }
 
 bool AST::CppPrettyPrinter::VisitLiteral(AST::Literal &node) {
   std::string value = std::get<std::string>(node.value.value);
   if (node.value.type != TT_CHARLIT && node.value.type != TT_STRINGLIT) {
+    if (node.value.type == TT_HEXLITERAL) {
+      print("0x");
+    }
```
JoshDreamland, line 96: Some test cases:
1. A hex string starting with 0x
2. A hex string starting with $
3. A decimal string like 123 (this should pass just fine)
4. A decimal string starting with 0 in GML mode
5. An octal string starting with 0o
6. An octal string starting with 0 in C++ mode.

I'm guessing this needs to be revised to look something like this:
```cpp
std::string_view cpplit = value;
switch (node.value.type) {
  case TT_HEXLITERAL: {
    print("0x");
    if (value.rfind("0x", 0) == 0) cpplit = cpplit.substr(2);
    else if (value[0] == '$') cpplit = cpplit.substr(1);
    break;
  }
  case TT_DECLITERAL: {
    while (cpplit.length() > 1 && cpplit[0] == '0') cpplit = cpplit.substr(1);
    break;
  }
  case TT_OCTLITERAL: {
    print("0");
    if (value.rfind("0o", 0) == 0) cpplit = cpplit.substr(2);
    break;
  }
}
```

CompilerSource/parsing/pretty_printer.cpp
JoshDreamland, line 605: Nit: string concatenation is more expensive than repeated stream-put operations. The system will run a little faster if we just `print(x); print(y); print(z);` instead of `print(x + y + z);`.

ENIGMAsystem/SHELL/Universal_System/estring.cpp
JoshDreamland: I think it's fine, sure. In C++ mode, `'A'` is a char literal with value 65. In GML mode, it's just another way of writing `"A"` (a string value). Those older games are pretty much all GML style.

JoshDreamland (commented):
Here's a batch of comments from me on the wedge of files GitHub feels like showing me right now. More recent changes I think. I will need to address two of them before submitting; if you don't want to tackle that, let me know.

samuelvenable:
I have another game to add to the list to test if you want, (it is a tad more complex than some of the other games listed):

https://samuel-venable.itch.io/key-to-success

In particular, the source code download marked as: `Source (ENIGMA Dev Environment for Desktop)`.

PR #2397, opened 2024-06-28
Google Summer of Code 2024 - Midterm Evaluation: Procedural Fragment Shader Generation Using Classic Machine Learning

k0T0z:
This is (hopefully) to finish #2390.

I will be updating my Logs inside my domain: [gsoc24-blog](https://k0t0z.github.io/gsoc24-blog/).

## TODO:

- [ ] Finish the generator.
- [ ] Find out how the shader code will be applied to a resource such as Sprites. 

# Operations implemented

# Others

Sometimes, Abseil, Protobuf, and gRPC packages are hard to link as some releases are incompatible. I have made [absl-proto-grpc-ci](https://github.com/k0T0z/absl-proto-grpc-ci) solution to address this issue. On Arch, if you get any link errors related to Abseil, Protobuf, or gRPC, you will have to build them manually and you can get the compatible versions from the included [README.md](https://github.com/k0T0z/absl-proto-grpc-ci/blob/master/README.md) file.

Issue #2390, opened 2024-05-01
Procedural Fragment Shader Generation Using Classic Machine Learning


k0T0z:
As mentioned in [A Tool for the Procedural Generation of Shaders using Interactive Evolutionary Algorithms
](https://arxiv.org/abs/2312.17587) paper, this project is divided into two major steps:

1. Implement the visual shader graph functionalities.
2. Apply a proper Genetic Algorithm to find the best shader for a pre-specified image.

The contributor may also read [Noise Modeler: An Interactive Editor and Library for Procedural Terrains via Continuous Generation and Compilation of GPU Shaders](https://link.springer.com/chapter/10.1007/978-3-319-24589-8_42) paper for a better understanding of the graph editor architecture.

PR #2399, opened 2024-08-09
Google Summer of Code 2024 - Final Evaluation: Procedural Fragment Shader Generation Using Classic Machine Learning

k0T0z:
This is (hopefully) to finish #2390.

Midterm work: #2397

I will be updating my Logs inside my domain: [gsoc24-blog](https://k0t0z.github.io/gsoc24-blog/).

### DIFF
--- a/CommandLine/emake/Makefile
+++ b/CommandLine/emake/Makefile
@@ -13,7 +13,7 @@ else
 endif
 
 CXXFLAGS  += -I../../CompilerSource -I$(PROTO_DIR) -I../libEGM -I../libEGM -I../../ENIGMAsystem/SHELL
-LDFLAGS   += $(OS_LIBS) -L../../ -lcompileEGMf -lEGM -lProtocols -lENIGMAShared -lgrpc++ -lprotobuf -lyaml-cpp -lpng 
+LDFLAGS   += $(OS_LIBS) -L../../ -lcompileEGMf -lEGM -lENIGMAShared -lProtocols -lgrpc++ -lgpr -lprotobuf -labseil_dll -lyaml-cpp -lpng
 
 ifeq ($(TESTS), TRUE)
 	TARGET=../../emake-tests
--- a/Config.mk
+++ b/Config.mk
@@ -60,6 +60,13 @@ ifeq ($(OS), DragonFly)
 	LDFLAGS  += -L/usr/local/lib
 endif
 
+# Which search priority to use for libraries (system or local)
+CUSTOM_LIB_SEARCH_PRIORITY := local
+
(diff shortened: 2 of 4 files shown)


shared/protos/Makefile
k0T0z, line 1: The changes in this file are to make it look more reasonable, it is the same as the old one.

Makefile
```diff
@@ -45,7 +45,7 @@ libEGM$(LIB_EXT): libEGM
 libEGM: .FORCE libProtocols$(LIB_EXT) libENIGMAShared$(LIB_EXT)
 	$(MAKE) -C CommandLine/libEGM/
 
-EMAKE_TARGETS = .FORCE ENIGMA libProtocols$(LIB_EXT) libEGM$(LIB_EXT)
```
k0T0z, line 48: No need for this target here as it is already inside ``$(LIB_PFX)compileEGMf$(LIB_EXT)``.

Config.mk
```diff
@@ -60,6 +60,13 @@ ifeq ($(OS), DragonFly)
 	LDFLAGS  += -L/usr/local/lib
 endif
 
+# Which search priority to use for libraries (system or local)
+CUSTOM_LIB_SEARCH_PRIORITY := local
```
k0T0z, line 64: This is a way to prioritize the local over the system libs as most developers will need to build grpc, protobuf, and absl from scratch if there is an incompatibility issue. Check the versions inside for what versions you need to clone: https://github.com/k0T0z/absl-proto-grpc-ci

CommandLine/emake/Makefile
```diff
@@ -13,7 +13,7 @@ else
 endif
 
 CXXFLAGS  += -I../../CompilerSource -I$(PROTO_DIR) -I../libEGM -I../libEGM -I../../ENIGMAsystem/SHELL
-LDFLAGS   += $(OS_LIBS) -L../../ -lcompileEGMf -lEGM -lProtocols -lENIGMAShared -lgrpc++ -lprotobuf -lyaml-cpp -lpng 
+LDFLAGS   += $(OS_LIBS) -L../../ -lcompileEGMf -lEGM -lENIGMAShared -lProtocols -lgrpc++ -lgpr -lprotobuf -labseil_dll -lyaml-cpp -lpng
```
k0T0z, line 16: This is to fix the link error in which it couldn't find an absl target.

shared/protos/treenode.proto
k0T0z, line 1: > R0bert — 11/10/2024 21:32
> @Saif but right now it doesn't matter since nobody is using RGM yet, it's not as big a deal

PR #2413, opened 2025-11-01
Fixed UB with widen() and narrow() Functions...

samuelvenable:
When an empty string is passed to these functions, before this pull request's changes were applied, it would result in undefined behavior.


samuelvenable:
WCHAR is a typedef of wchar_t last I checked, or it may even be a macro because it's in all caps.

PR #2415, opened 2025-12-08
ai port to protobutt 3

fundies:
p1 of making rgm build on my system again


CommandLine/libEGM/gmk.cpp
```diff
@@ -885,17 +909,34 @@ int LoadActions(Decoder &dec, std::string* code, std::string eventName) {
...
-        { ArgumentKind::ARG_SOUND,      {&Argument::mutable_sound,       TypeCase::kSound      }},
-        { ArgumentKind::ARG_BACKGROUND, {&Argument::mutable_background,  TypeCase::kBackground }},
-        { ArgumentKind::ARG_SPRITE,     {&Argument::mutable_sprite,      TypeCase::kSprite     }},
-        { ArgumentKind::ARG_SCRIPT,     {&Argument::mutable_script,      TypeCase::kScript     }},
-        { ArgumentKind::ARG_FONT,       {&Argument::mutable_font,        TypeCase::kFont       }},
-        { ArgumentKind::ARG_OBJECT,     {&Argument::mutable_object,      TypeCase::kObject     }},
-        { ArgumentKind::ARG_TIMELINE,   {&Argument::mutable_timeline,    TypeCase::kTimeline   }},
-        { ArgumentKind::ARG_ROOM,       {&Argument::mutable_room,        TypeCase::kRoom       }},
-        { ArgumentKind::ARG_PATH,       {&Argument::mutable_path,        TypeCase::kPath       }}
+        { ArgumentKind::ARG_SOUND,      {[getMutableString](Argument* a) { return getMutableString(a, "sound"); },       TypeCase::kSound      }},
```
JoshDreamland, line 931: The longer I stare at this, the less I understand what happened, here. So, above, it just hallucinates this function "`getMutableString`" that returns a string which "magically" has a setter in it...? Except the "setter" it creates is just a thread_local vector, which isn't hooked up to anything, and it returns a normal string pointer... which I think it also leaks, since the setter is never called. And by the way, these aren't really string fields (they are technically string on the wire because they are messages, but string* is not the correct return type...).

I think the AI got confused because MutatorMap treats all these protobuf return types as string*... I'm not sure that's correct. Isn't each one a separate message type? It may need to treat them as void* and do the casting elsewhere.

CommandLine/libEGM/egm-events.cpp
```diff
@@ -42,7 +42,12 @@ void LoadObjectEvents(const fs::path& fPath, google::protobuf::Message *m, const
 
 void WriteObjectEvents(const fs::path& fPath, const google::protobuf::RepeatedPtrField<buffers::resources::Object_EgmEvent>& events, const EventData* event_data) {
   for (auto &e : events) {
-    auto event = event_data->get_event(e.id(), {e.arguments().begin(), e.arguments().end()});
+    std::vector<std::string> args;
+    args.reserve(e.arguments_size());
+    for (const auto& arg : e.arguments()) {
+      args.push_back(std::string(arg));
+    }
+    auto event = event_data->get_event(std::string(e.id()), args);
```
JoshDreamland, line 50: I don't think this change is necessary. It's possible the vector argument needs to be constructed explicitly now, though it is not marked explicit by the standard afaik.

I guess `e.id()` *does* need to be wrapped in `string()`, but I think it's better (as in most cases in this PR) to just modify `get_event()` to accept a `string_view`.

PR #2423, opened 2026-05-21
New ci & fixes josh dropped cause hes annoying

fundies:


JoshDreamland (approved):
Windows macro guard is a bit ugly, but I guess that's a Microsoft problem... rest looks good; thanks!

